PR #5, opened 2022-01-06
introduce jest tests

cogoo:
- build(jest): setup jest
- test(parseurl): add tests for parseURL function


js/package.json
```diff
@@ -27,6 +27,7 @@
         "build": "yarn clean && tsc -p tsconfig.json && tsc -p tsconfig-cjs.json",
         "postbuild": "echo '{\"type\":\"commonjs\"}' | npx json > lib/cjs/package.json && echo '{\"type\":\"module\"} ' | npx json > lib/esm/package.json",
         "test": "node --experimental-specifier-resolution=node --loader ts-node/esm test/index.ts",
+        "test:jest": "jest",
```
jordansexton, line 30: This can just be the "test" script, the test/index.ts file needs to be replaced with proper tests anyway. We can just move the file and rename the script that runs it for now.

js/jest.config.js
```diff
@@ -0,0 +1,5 @@
+/** @type {import('ts-jest/dist/types').InitialOptionsTsJest} */
+export default {
+  preset: 'ts-jest',
+  testEnvironment: 'node',
+};
```
jordansexton, line 5: Can this file just be JSON?

js/src/wallet/parseURL.test.ts
```diff
@@ -0,0 +1,92 @@
...
+        it('throws an error on invalid recepient', () => {
+            const url = 'solana:0xffff';
+            expect(() => parseURL(url)).toThrow('recipient invalid');
+        });
+
+        it('throws an error on missing amount', () => {
+            const url = 'solana:mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN';
+
+            expect(() => parseURL(url)).toThrow('amount missing');
+        });
```
jordansexton, line 60: This isn't a problem with this test, but we should probably support payment links that don't specify an amount, and the wallet should allow input. This would allow you to use the protocol for transfers between friends like a CashApp QR code.

js/src/wallet/parseURL.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+            expect(() => parseURL(url)).toThrow('amount NaN');
+        });
+
+        it('throws an error on zero amount', () => {
+            const url = 'solana:mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN?amount=0';
+
+            expect(() => parseURL(url)).toThrow('amount zero');
+        });
+
```
jordansexton, line 79: Could use a test for a negative amount.

js/test/index.ts
```diff
@@ -30,6 +30,7 @@ import { createTransaction, parseURL } from '../src/wallet';
         '&memo=OrderId5678';
 
     const originalURL = NATIVE_URL;
+    console.log(originalURL);
```
jordansexton, line 33: We can move this file since it's not a real test.

jordansexton (commented):
Looks great! This will help a lot and serves as a basis for what to document.

js/src/wallet/parseURL.test.ts
cogoo: That is very clever

js/test/index.ts
cogoo: I was thinking of converting it into a test. thoughts?

or keep it as is and move it to `/example` directory?

jordansexton: I think we can keep it as is for now, extract the behavior to tests, and decide what to do with it later.

js/src/wallet/parseURL.test.ts
jordansexton: Sounds like this is supported by FTX so we should make it part of the spec / impl.

cogoo: nice!

Potentially a different PR for that

jordansexton: Yeah sounds good.

js/src/wallet/parseURL.test.ts
```diff
@@ -0,0 +1,92 @@
...
+            expect(() => parseURL(url)).toThrow('recipient invalid');
+        });
+
+        it('throws an error on missing amount', () => {
+            const url = 'solana:mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN';
+
+            expect(() => parseURL(url)).toThrow('amount missing');
+        });
+
+        it.each([['1milliondollars'], [-0.1], [-100]])('throws an error on invalid amount: %p', (amount) => {
```
jordansexton, line 62: Nice!

js/src/wallet/parseURL.test.ts
```diff
@@ -0,0 +1,92 @@
...
+            it('should parse successfully', () => {
+                const url =
+                    'solana:mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN?amount=0.01&reference=82ZJ7nbGpixjeDCmEhUcmwXYfvurzAgGdtSMuHnUgyny&label=Michael&message=Thanks%20for%20all%20the%20fish&memo=OrderId5678';
+
+                const parsed = parseURL(url);
+
+                expect(parsed.recipient).toEqual(new PublicKey('mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN'));
+                expect(parsed.amount).toEqual(new BigNumber(0.01));
+                expect(parsed.token).toBeUndefined();
+                expect(parsed.references).toEqual([new PublicKey('82ZJ7nbGpixjeDCmEhUcmwXYfvurzAgGdtSMuHnUgyny')]);
```
jordansexton, line 17: Hmm this is interesting that this works -- how are the two `PublicKey` instances tested for equality here?

cogoo: `.toEqual` does a deep comparison. Loops through each key and compares the values

jordansexton: Got it, I just didn't realize `toEqual` automatically called that. Thanks!

cogoo: > Use .toEqual when you want to check that two objects have the same value. This matcher recursively checks the equality of all fields, rather than checking for object identity—this is also known as "deep equal"

PR #7, opened 2022-01-07
test/create transaction

cogoo:
- test(createtransaction): general error checks
- test(createtransaction): error checks for transferring native SOL
- test(create transaction): validating transaction

### DIFF
--- a/core/test/createTransaction.test.ts
+++ b/core/test/createTransaction.test.ts
@@ -1,11 +1,11 @@
-import { clusterApiUrl, Connection, Keypair, PublicKey } from '@solana/web3.js';
+import { clusterApiUrl, Connection, Keypair, LAMPORTS_PER_SOL, PublicKey, SystemProgram } from '@solana/web3.js';
 import BigNumber from 'bignumber.js';
-import { createTransaction, CreateTransactionError } from '../../src/wallet/createTransaction';
+import { createTransaction } from '../../src/wallet/createTransaction';
 
 describe('createTransaction', () => {
     let connection: Connection;
 
-    beforeEach(() => {
+    beforeAll(() => {
         const endpoint = clusterApiUrl('devnet');
         connection = new Connection(endpoint, 'confirmed');
     });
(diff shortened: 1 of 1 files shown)


js/test/wallet/createTransaction.test.ts
```diff
@@ -0,0 +1,194 @@
...
+                const transaction = await createTransaction(
+                    connection,
+                    wallet.publicKey,
+                    new PublicKey('mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN'),
+                    new BigNumber(0.01),
+                    {}
+                );
+
+                expect(transaction.instructions).toHaveLength(1);
+                // FIXME: Best way to validate?
```
cogoo, line 33: @jordansexton unsure on the best way to validate the correct transaction was created

js/test/wallet/createTransaction.test.ts
```diff
@@ -0,0 +1,194 @@
...
+                        {
+                            token: new PublicKey('8FRFC6MoGGkMFQwngccyu69VnYbzykGeez7ignHVAFSN'),
+                        }
+                    );
+                }).rejects.toThrow('amount decimals invalid');
+            });
+
+            // FIXME: can't get past this error
+            // Fails at getting associated token account
+            it.skip('throws an error on unintialized payer', async () => {
```
cogoo, line 139: @jordansexton for the implementation, we may need to consider having a getOrCreateAssociatedTokenAccount

js/test/wallet/createTransaction.test.ts
jordansexton: One quick way to do it would be to create the expected transaction and serialize it, and then compare the serialized bytes / base58 encoding.

js/test/wallet/createTransaction.test.ts
```diff
@@ -0,0 +1,194 @@
...
+                    connection,
+                    wallet.publicKey,
+                    new PublicKey('mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN'),
+                    new BigNumber(0.01),
+                    {
+                        memo: 'Thanks for all the fish',
+                    }
+                );
+
+                expect(transaction.instructions).toHaveLength(2);
```
jordansexton, line 49: Same thing here, we could just serialize the tx.

js/test/wallet/createTransaction.test.ts
jordansexton: No, the payer and recipient need to exist first, we don't need to create them. It's up to wallets to ensure they do (but you can't send USDC unless you already have the ATA, so this is moot). For testing purposes, we can just require that you run the test validator and create some accounts on localnet in the setup I imagine?

js/test/wallet/createTransaction.test.ts
```diff
@@ -0,0 +1,194 @@
...
+                }).rejects.toThrow('payer not initialized');
+            });
+
+            it.todo('throws an error on frozen payer');
+
+            it.todo('throws an error on unitialized recipient');
+
+            it.todo('throws an error on frozen recipient');
+
+            it.todo('throws an error on insufficient funds');
```
jordansexton, line 161: Actually, I wonder if we should try to mock the return values of `getAccount` etc. for these and for the checks above.

js/test/wallet/createTransaction.test.ts
cogoo: hmm, I guess the flow would be for SPL:
1. Wallet parses the URL
2. Gets the token and creates the ATA if it doesn't exist
3. Then creates the transaction

js/test/wallet/createTransaction.test.ts
cogoo: Yup, that's what I was leaning toward for these ones

js/test/wallet/createTransaction.test.ts
jordansexton: No, not quite. 2 is what doesn't make sense -- a user can't send, say, USDC, unless they already have an ATA for USDC. A wallet doesn't need to create an ATA for you (you either have it because you have the tokens or you don't).

js/test/wallet/createTransaction.test.ts
jordansexton: I'm thinking we should maybe do this for anything that needs the SPL token functions. It's better if the tests don't require running a local validator / devnet

js/test/wallet/createTransaction.test.ts
cogoo: ahh sorry yes! ofcourse.

js/test/wallet/createTransaction.test.ts
jordansexton: I think we should decide on a single approach either way -- mock everything or nothing.

Argument for: things like `connection.getTransaction` in `validateTransactionSignature` might be easier to mock than to create the necessary state on chain. Tests will be faster and don't require running a local validator.

Argument against: if we mock it and there are breaking changes, we don't discover them, and it's also annoying to mock things.

cogoo: yup, makes sense

cogoo: I think for these unit tests, we mock them. We can have one integration test that will catch these breaking changes

jordansexton: Sounds good. Let's mock everything then! We can worry about integration tests after

js/test/wallet/createTransaction.test.ts
```diff
@@ -0,0 +1,436 @@
...
+
+describe('createTransaction', () => {
+    let connection: Connection;
+    let wallet: Keypair;
+    const recipientPublicKey = new PublicKey('mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN');
+    const accountOwnedByWallet = Keypair.generate();
+    const programAccount = Keypair.generate();
+
+    beforeAll(async () => {
+        wallet = Keypair.generate();
```
jordansexton, line 32: We can use random keys but they need to be generated outside of the test and provided statically so they will be deterministic across test runs.

js/test/wallet/createTransaction.test.ts
```diff
@@ -0,0 +1,436 @@
...
+                        executable: false,
+                        owner: wallet.publicKey,
+                        lamports: LAMPORTS_PER_SOL / 100,
+                    },
+                    [programAccount.publicKey.toBase58()]: {
+                        executable: true,
+                        owner: SystemProgram.programId,
+                        lamports: LAMPORTS_PER_SOL / 100,
+                    },
+                };
```
jordansexton, line 62: This gets created every time we call `getAccountInfo`. Shouldn't we just create it once?

js/test/wallet/createTransaction.test.ts
```diff
@@ -0,0 +1,436 @@
...
+    {
+        token,
+        references,
+        memo,
+    }: {
+        token?: PublicKey;
+        references?: PublicKey[];
+        memo?: string;
+    }
+): Transaction {
```
jordansexton, line 409: This helper function merely duplicates the entire logic between the test code and the source code. This isn't really testing what we want.

It's better to just serialize the expected transaction (as JSON or bytes or base58 or something) to deterministically test for regression.

cogoo:
Need to rethink the test for transaction. Since `connection` is mocked, not able to call `serialize` method to compare the transaction

PR #8, opened 2022-01-08
test(encode url): test for encode url

cogoo:
tests for encodeURL


cogoo:
Will circle back and change the assertions. With the latest refactor, the current tests are too tied to the order of the params.

jordansexton:
> the current tests are too tied to the order of the params.

Not sure I understand. Even though they are all optional now, the order of the params is still deterministic in the function output.

cogoo:
Adding any new params anywhere but the bottom will cause the tests to fail. The test shouldn't really care about the order.

Unless you actually want to enforce the order? (Can't think of a reason why)

PR #9, opened 2022-01-08
chore(validate transaction signature): scaffold test

cogoo:
test for validate transaction signature


cogoo:
Will re-open after writing tests

PR #10, opened 2022-01-08
fix(parse url): remove check for missing amount

cogoo:


core/test/parseURL.test.ts
```diff
@@ -34,6 +34,21 @@ describe('parseURL', () => {
...
+            it('should parse without an amount', () => {
+                const url =
+                    'solana:mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN?reference=82ZJ7nbGpixjeDCmEhUcmwXYfvurzAgGdtSMuHnUgyny&label=Michael&message=Thanks%20for%20all%20the%20fish&memo=OrderId5678';
+
+                const parsed = parseURL(url);
+
+                expect(parsed.amount).toBeUndefined();
+                expect(parsed.recipient).toEqual(new PublicKey('mvines9iiHiQTysrwkJjGf2gb9Ex9jXJX8ns3qwf2kN'));
+                expect(parsed.token).toBeUndefined();
+                expect(parsed.reference).toEqual([new PublicKey('82ZJ7nbGpixjeDCmEhUcmwXYfvurzAgGdtSMuHnUgyny')]);
```
jordansexton, line 47: How does `toEqual` work here for PublicKey instances? Don't we need to compare their string base58 representations?

PR #12, opened 2022-01-13
Add spec draft

jordansexton:
Formatted version viewable @ https://github.com/solana-labs/solana-pay/tree/spec/spec

### DIFF
--- a/spec/README.md
+++ b/spec/README.md
@@ -1,0 +1,121 @@
+# SolanaPay Specification Draft
+
+This is a draft specification of the SolanaPay URL protocol.
+
+## v0.1 Base Protocol
+
+This is the original proposal, compiled and adapted from discussion on https://github.com/solana-labs/solana/issues/19535.
+
+Rough consensus on this spec has been reached, and WIP implementations exist in Phantom, FTX, and Slope.
+
+---
+
+A standard URI scheme across wallets for native SOL and SPL Token payments (transfers) is desirable.  This scheme for example could be encoded as a QR Code.
+
+TrustWallet has taken the lead with `solana:<ADDRESS>?amount=<AMOUNT>`, which is essentially [BIP 21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) but with the `bitcoin:` scheme replaced with `solana:`.  This is a great start.
+
+The existing BIP 21 `label=string` ("Label for that address (e.g. name of receiver)") and `message=string` ("message that describes the transaction to the user ") fields are directly relevant. This labels are informative only and are not encoded into the on-chain transaction
+
+The `amount` field is always interpreted to be a decimal number of "user" units. For SOL, that's SOL and not lamports.  For tokens,`uiAmountString` and not `amount` (reference: [Token Balances Structure](https://docs.solana.com/developing/clients/jsonrpc-api#token-balances-structure)).  If the provided decimal fractions exceed what's supported for SOL (9) or the token (mint specific), the URL must be considered malformed URL and rejected. Scientific notation is prohibited. If the amount is not provided, the wallet should prompt the user for the amount.
+
+For SPL Token transfers, an additional `spl-token=<MINT_ADDRESS>`, is required to define the token type.  If no `spl-token=` field is specified, the URL describes a native SOL transfer. For SPL Token transfers, the [Associated Token Account](https://spl.solana.com/associated-token-account) convention must be used.  Transfers to auxiliary token accounts are not supported.
+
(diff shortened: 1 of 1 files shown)


spec/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+This proposal is to add an optional `request=<URL_ENCODED_URL>` parameter included in the link.
+
+An interactive protocol between the merchant and wallet follows:
+
+1. The merchant presents a QR code with the following payment link:
+```
+solana:<ADDRESS>?request=https%3A%2F%2Fmerchant.com%2Fsolanapay
+```
+Any of the parameters of the spec can also be included. `<ADDRESS>` could be considered optional. Perhaps an invalid address (e.g. `a`, `x`, or `_`) could be provided for compatibility. Regardless, it should not be used by the wallet if `request` is provided.
```
jordansexton, line 91: I think we should allow URLs with no recipient of the form
```
solana:?request=https://...
```
This is shorter (will encode in a QR code better) and is more semantically correct than providing a value that shouldn't be used and hoping it isn't.

jordansexton:
There's a branch with a reference implementation for the proposed `request` parameter:

- [Encode a payment URL](https://github.com/solana-labs/solana-pay/blob/request/js/src/encodeURL.ts)
- [Parse a payment URL](https://github.com/solana-labs/solana-pay/blob/request/js/src/parseURL.ts)
- [Request a transaction](https://github.com/solana-labs/solana-pay/blob/request/js/src/requestTransaction.ts)

spec/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+2. The customer scans the QR code and opens their wallet app.
+
+3. The wallet parses link and prompts the user to make a request to `https://merchant.com/solanapay`.
+
+This is analogous to connecting a wallet to a dapp, so obtaining the user's permission is important for privacy.
+
+6. If permitted, the wallet makes a request to
+```
+https://merchant.com/solanapay?from=<WALLET_ADDRESS>&<...PARAMs>
```
jordansexton, line 101: This doesn't strictly need to be `https://` links -- I think two mobile apps could communicate with each with deep links if non-HTTPS schemes are supported, e.g.
```
merchant-app:custom-deep-link-format?from=<WALLET_ADDRESS>
```

spec/README.md
```diff
@@ -0,0 +1,121 @@
...
+```
+The wallet should include any parameters from the URL provided, except for the `request` parameter.
+
+5. The merchant responds with a JSON object:
+```
+{"transaction":"<TRANSACTION>"}
+```
+The `transaction` field must be a base58-encoded [serialized transaction](https://solana-labs.github.io/solana-web3.js/classes/Transaction.html#serialize).
```
jordansexton, line 109: Rationale for Base58 here is that it's included as a dependency of `@solana/web3.js` so it should be available in many contexts you need to work with `Transaction` objects. This may be a poor rationale for mobile wallets though. Base64 or another encoding could work.

mvines: Base64 is my strong preference for any new work.  Base58 is such a performance landmine, and base64 support is ubiquitous.

spec/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+We propose adding an optional `reference=<REFERENCE_ADDRESS>` parameter. If provided, a unique account address should be generated and included as a value. 
+
+If `reference` parameters are included, the wallet should attach them as read-only, non-signer keys to the `SystemProgram.transfer` or `TokenProgram.transfer` instruction.
+
+This allows any transaction using the `reference` address to be located on chain with the [`getSignaturesForAddress`](https://docs.solana.com/developing/clients/jsonrpc-api#getsignaturesforaddress) RPC method.
+
+Multiple `reference` parameters should be supported if the requester wishes to categorize transactions on chain.
+
+## v0.3 Request Link
```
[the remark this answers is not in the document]
benmills: This is very cool. While I agree there are some considerations merchants/dapps would need to make for merchants to co-sign transactions the utility that provides seems massive. Could you imagine your gas relay concept building on top of this? Basically Octane uses the Request Link specification to co-sign transactions and cover gas fees. The Octane protocol could develop ways to prevent spam on behalf of all merchants letting them get the benefit without having to deal with the complexity.

spec/README.md
benmills: I don't know if I follow this. Is the issue with providing the address that the transaction will be co-signed making it unclear what the address is used for?

jordansexton: The semantic issue is that there may be no address to provide. When using the `request` parameter, the transaction can be anything -- it doesn't even need to be a payment. It could be a transaction to receive a gift or invitation from the merchant for scanning a wallet. Even if it is a payment, the receiving address could be dynamic, or a smart contract, etc.

spec/README.md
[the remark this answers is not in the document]
jordansexton: Yeah, I think merchants could run a relayer that only sponsors their customers transactions, or a service could run a relayer for them. GaaS

jordansexton:
Superseded by #25 and #26

PR #17, opened 2022-01-20
docs(root readme): root level readme

cogoo:
add root level readme


jordansexton:
@cogoo can this be closed? I think @benmills recent PR covers this, but let me know if there's something here we should merge.

PR #23, opened 2022-01-26
docs(core): Update qr code to be smaller

jacobcreech:
**Problem**: QR code took up a lot of screen real estate

**Solution**: Made it smaller in the md.


jordansexton:
Hmm would it be possible to actually change the image size to be smaller? It's probably not exactly square right now because of the "Powered by Solana Pay" line, so this change is causing it to look skewed.

jacobcreech:
@jordansexton sure thing. I'll match the aspect ratio. Do you actually want the image smaller than 300px width, or just the right ratio?

jordansexton:
Just the right ratio, but I think 300 is okay (whatever looks okay for desktop and mobile I guess)

PR #24, opened 2022-01-26
docs(core): Update docs format + payment url scheme reference

jacobcreech:
Addressed:

- Giving full payment URL scheme and definition
- Reusing table to explain URL scheme
- Note on confirmation for finding transaction signatures
- Fixing `amountInLamports` retrieval to match full source

### DIFF
--- a/core/README.md
+++ b/core/README.md
@@ -138,15 +138,21 @@ See [full code snippet][6]
 
 <br>
 
-The `recipient` must be a native SOL address. So, for our merchant example, the recipient is the merchant's native SOL wallet address.
+The full scheme:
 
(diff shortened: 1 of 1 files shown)


cogoo:
The previous iteration of the doc used the tabular format. It was stripped out in favour of having a typedoc that we'll link to for the sake of brevity of the readme.

Thoughts @jordansexton

jordansexton:
#25 included [the spec](https://github.com/solana-labs/solana-pay/blob/master/SPEC.md). I think some of these edits make sense but we should DRY things out, wdyt?

jacobcreech:
@jordansexton I think it makes more sense to link the spec and remove the tables altogether. Spec is much better covering what each parameter does. I'll update

PR #27, opened 2022-01-26
Updated the minimum Node.js version

SunitRoy2703:
`yarn install` gives an error saying "Expected version ">=14.17.0". Got "14.16.0"

Observed while running the [point-of-sale in local system](https://github.com/solana-labs/solana-pay/tree/master/point-of-sale#getting-started)

## Reproduction
1. `nvm use 14.16.0`
2. `yarn install`
3. Error: 
```properties
error @toruslabs/solana-embed@0.0.9: The engine "node" is incompatible with this module. Expected version ">=14.17.0". Got "14.16.0"
error Found incompatible module. 
```
4. `nvm use 14.17.0`
5. `yarn install`
6. builds correctly

### DIFF
--- a/point-of-sale/README.md
+++ b/point-of-sale/README.md
@@ -8,7 +8,7 @@ You can use the code as a reference or run it yourself to start accepting decent
 
 To build and run this app locally, you'll need:
 
--   Node.js v14+
+-   Node.js v14.17.0 or above
 -   Yarn
 -   <details>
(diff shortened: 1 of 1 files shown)


jordansexton:
Hmm, this sounds like an issue we should take up with Torus.

@metallicalfa @ArpitAgrawal321 is there a reason the Torus Embed lib requires this specific minor version of Node 14 or higher?

jordansexton:
Thanks @SunitRoy2703. Merged to fix the issue for now, but perhaps Torus will be able to set a lower version and we can revert.

PR #42, opened 2022-02-01
update cd to 'solana-pay' for Github CLI

jcstein:
This is an update to the `cd` command for the Github CLI version

### DIFF
--- a/point-of-sale/README.md
+++ b/point-of-sale/README.md
@@ -64,7 +64,7 @@ gh repo clone solana-labs/solana-pay
 Install dependencies for point-of-sale
 
 ```shell
-cd solana-pay
+cd solana-pay && cd point-of-sale
 yarn install
 ```
 
--- a/point-of-sale/README.md
+++ b/point-of-sale/README.md
@@ -48,6 +48,13 @@ Clone the repository
 git clone https://github.com/solana-labs/solana-pay.git
 ```
 
+Install dependencies for point-of-sale
+
+```shell
+cd point-of-sale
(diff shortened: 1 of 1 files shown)


jordansexton:
Thanks!

jcstein:
You're welcome! Thanks for cleaning it up 😊

jcstein:
Any idea what I may be doing wrong here? Thank you!

scrot 2022-02-01 at 10 22 15 PM
!

jordansexton:
@jcstein I've seen this kind of thing happen if you end up with two different versions of some dependency. What I would do is run `yarn nuke` and then `yarn install` from the point-of-sale directory to see if it fixes it.

jcstein:
thank you @jordansexton! This makes sense now. I think what (may have) happened is that I ran `yarn install` in the `solana-pay` directory and then again in `point-of-sale` more than once after I ran into this issue.

jordansexton:
Yeahhh, I think it might be good for us to set up yarn workspaces or something if we're going to continue to use a monorepo so that dependencies get properly linked / deduped.

jcstein:
sounds good, thanks again. I'm still learning, but this makes sense. please forgive me if I am asking too much 😅 . I've got things running successfully, what would be the next steps to set myself up to test out a transaction?

scrot 2022-02-01 at 11 01 34 PM

jcstein:
got it running! gn ✌️ and thanks again @jordansexton

jordansexton:
Glad you got it working!

PR #51, opened 2022-02-03
README - link to Shopify reference implementation

intelliot:
This was a bit difficult to find since it wasn't linked from the README anymore


jordansexton:
Good catch, thanks!

PR #77, opened 2022-02-12
Specify and implement Solana Pay Transaction Requests

jordansexton:
Rationale is covered in https://github.com/solana-labs/solana-pay/issues/26.

This PR specifies and implements Transaction Requests as a new format. See https://github.com/solana-labs/solana-pay/blob/link-request/SPEC.md#specification-transaction-request for the spec.

This PR also substantially refactors the Transfer Request implementation (but the spec is compatible).

Todo
- [ ] Add tests
- [ ] More docs
- [ ] Update point-of-sale app


core/src/createTransaction.ts
```diff
@@ -148,3 +90,78 @@ export async function createTransaction(
 
     return transaction;
 }
+
+async function createSystemInstruction(
+    connection: Connection,
+    payer: PublicKey,
+    recipient: PublicKey,
+    amount: BigNumber
+): Promise<TransactionInstruction> {
```
mcintyre94, line 99: Am I right that the newly supported use cases for things like bidirectional NFTs would require the server to build a `Transaction` without using `createTransaction`, or to add extra instructions after creating a transaction using it? If so, would it be worth exporting these `createSystemInstruction` and `createSPLTokenInstruction` functions as building blocks that can be used to make creating complex transactions a bit easier?

core/src/encodeTransactionRequestURL.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    if (message) {
+        params.push(['message', message]);
+    }
+
+    if (params.length) {
+        url += '?' + params.map(([key, value]) => key + '=' + encodeURIComponent(value)).join('&');
+    }
+
+    return url;
```
mcintyre94, line 38: Nit: this would be cleaner using [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

```
const params = new URLSearchParams();

if(label) {
    params.append('label', label);
}

if(message) {
    params.append('message', message)
}

if(Array.from(params).length) {
    url += '?' + params.toString();
}

return url;
```

Disregard if we support IE!

core/src/encodeTransferRequestURL.ts
```diff
@@ -0,0 +1,77 @@
...
+        params.push(['message', message]);
+    }
+
+    if (memo) {
+        params.push(['memo', memo]);
+    }
+
+    if (params.length) {
+        url += '?' + params.map(([key, value]) => key + '=' + encodeURIComponent(value)).join('&');
+    }
```
mcintyre94, line 74: As above re `URLSearchParams` if possible :)

core/src/findReference.ts
```diff
@@ -1,41 +1,42 @@
...
-    if (!length) throw new FindTransactionSignatureError('not found');
+    if (!length) throw new FindReferenceError('not found');
 
     // If multiple transaction signatures are found within the limit, return the oldest one.
     if (length < (options?.limit || 1000)) return signatures[length - 1];
 
     try {
         // In the unlikely event that more signatures than the limit are found, recursively find the oldest one.
-        return await findTransactionSignature(connection, reference, options, finality);
+        return await findReference(connection, reference, options, finality);
```
mcintyre94, line 37: Is this an infinite loop since the params are identical? In particular should we be updating `options.before` to the latest signature seen? Basing on https://solana-labs.github.io/solana-web3.js/modules.html#SignaturesForAddressOptions

core/src/parseURL.ts
```diff
@@ -10,43 +12,76 @@ export class ParseURLError extends Error {
...
+    if (typeof url === 'string') {
+        if (url.length > 2048) throw new ParseURLError('length invalid');
+        url = new URL(url);
+    }
+
+    const { protocol, pathname } = url;
+    if (protocol !== SOLANA_PROTOCOL) throw new ParseURLError('protocol invalid');
+    if (!pathname) throw new ParseURLError('pathname missing');
+
+    return /[^A-Za-z0-9]/.test(pathname) ? parseTransactionRequestURL(url) : parseTransferRequestURL(url);
```
mcintyre94, line 63: Question: does this work because `pathname` for a transaction request is an account, where for a transfer request it's a URL and must have some non-alphanumeric characters?

core/src/parseURL.ts
```diff
@@ -10,43 +12,76 @@ export class ParseURLError extends Error {
...
+}
+
+function parseTransactionRequestURL({ pathname, searchParams }: URL): TransactionRequestURL {
+    const link = new URL(decodeURIComponent(pathname));
+    if (link.protocol !== HTTPS_PROTOCOL) throw new ParseURLError('link invalid');
```
mcintyre94, line 68: Question: would you consider supporting HTTP too for ease of localhost dev?

core/src/encodeTransactionRequestURL.ts
jordansexton: Good suggestion, done.

core/src/encodeTransferRequestURL.ts
jordansexton: Done.

core/src/parseURL.ts
jordansexton: Nope. The protocol specifies that the link must be rejected as malformed unless it is an absolute HTTPS link.

Using HTTPS prevents MITM attacks that could change the transaction a user signs, which is critically important.

We should provide documentation on running a simple local HTTPS proxy instead.

core/src/findReference.ts
jordansexton: Good catch! I'll fix this.

jordansexton: Done.

core/src/createTransaction.ts
jordansexton: In terms of usage, I think it's binary -- you're either using a transfer request, and the SDK provides an opinionated framework, or you're using a transaction request, and the implementation is entirely up to you.

Patterns will emerge, but should be encouraged outside of the SDK. It's only intended to be a reference implementation of the specification, and the spec doesn't care how you create arbitrary transactions.

core/src/parseURL.ts
jordansexton: Refactored to `/[:%]/` since it's simpler, and if `.sol` or something becomes supported in the future, this won't interpret it as a transaction request URL.

jordansexton:
@PaulFidika 

> What is the rationale behind URI-encoding any URL query-parameters?

1. `label` and `message` are query params that are part of the spec. If a `link` used those, there would be no way to tell if they were intended for the wallet to render or intended for the `link` to receive.
2. Other query parameters may be added to the spec in the future. Therefore URL encoding a `link` with query parameters future proofs the `link`.

In general, merchants/service providers should prefer links without query parameters since they will encode in fewer characters.

> Also does this specification allow for requesting signatures for the purpose of authentication? As in window.solana.signMessage()

No, this isn't a design goal. Transaction request URLs are to request a transaction that can be broadcast on Solana.

Transfer request URLs work similarly. One way or another, a transaction needs to happen on chain for the merchant to find out about it.

This could be added by future specification though! Please open an issue that describes how you see the interaction between the point of sale, merchant, wallet, and user working.

PaulFidika:
@jordansexton 
Okay got it--the URL-encoding avoids query-param name collisions. It makes sense to separate the two (params meant for the wallet and params meant for the application).

Was the primary motivation for using URLs in the first place for Solana Pay compactness? I played with the idea of passing transactions as base64 encoded JSON, but the URLs were more compact every time.

I was thinking message-signing would be useful for authentication; for example, a desktop app might prompt me to scan a QR code with my Phantom-Wallet app on my phone, prompting me to sign a message which proves ownership of my Pubkey, then logging me into the desktop app.

I'll think that through a little more and then open an issue detailing my thoughts.

jordansexton:
> Was the primary motivation for using URLs in the first place for Solana Pay compactness?

No, compactness helps, but by far the most important thing for UX is that they work as deep links. A user scanning with a camera app on their phone or using an NFC tag will be prompted to launch the link in a wallet that registers itself as a handler of `solana:` links.

> I was thinking message-signing would be useful for authentication.

I agree, this could be useful. You'd need define a way for wallets to send that signed message somewhere and for the merchant to receive it.

One way to think about Solana Pay is that it allows an application and a mobile wallet to use the chain itself to talk to each other. Transaction requests augment that, using an intermediate request to a remote server.

We could potentially add another type that makes a signing/encryption request, and the wallet would send a request to a remote server with the signed/encrypted message.

It would be important to make sure that the message being signed couldn't be a transaction message, but I think that's on the wallet to handle.

PaulFidika:
Thanks for the clarification and suggestions! That was really helpful.

> We could potentially add another type that makes a signing/encryption request, and the wallet would send a request to a remote server with the signed/encrypted message.

Yes--exactly this. I'm calling this proposal 'Solana Auth' for now. Although it's not unique to Solana--this authentication method could be done with any app that controls a user's private key.

How long do you think it will take Phantom Wallet to integrate your new Transaction-Request spec and release it into production? Are we talking like weeks or months from now? I'll get in contact with their dev team this week.

jordansexton:
~2-3 weeks. Already working on it.

mcintyre94:
Spec suggestion: what about allowing an updated `message` param to be returned optionally alongside the transaction in response to a transaction request?

The interaction model I'm thinking of is something like:

- A QR code embeds the transaction request to `solana:https://api.callumscookies.com...?label=callums cookies&message=6 cookies ($5)`
- The wallet displays the label + message, user confirms to initiate the API request
- The API sees that the account is a loyal customer and returns:
```
{
   "transaction": "...", // some transaction with amount $4.50
   "message": "6 cookies (10% off)"
}
```
- The wallet displays the previous label + new message + asks user to confirm the transaction

This would allow a vendor to give an indication of what's special about the transaction since it's account-aware, eg. a discount that has been applied or maybe "You'll receive an NFT of your new  shoes" or whatever specific features they're using for their transaction. It'd also make it less confusing if you indicated a price in the initial message (since it doesn't have an amount that might be nice to do), and then the account-aware transaction was a different price, having eg. the discount in the updated message would make it a bit clearer.

Ajayneueworld:
Is there a reason why I can't see the confirmation  after making the payment on screen ? I have downloaded the code from github and using solana sale-of-point as instructed..

PaulFidika:
> Spec suggestion: what about allowing an updated `message` param to be returned optionally alongside the transaction in response to a transaction request?

Yes exactly; displaying a message in the user's wallet (along with the origin of the request) could be helpful. Although the message could be used by malicious sites to confuse the user about what they're buying or what the transaction does. Ideally by the time the user gets to the payment screen in their wallet they should already understand what they're doing.

I included a 'message' param in my spec here:

https://github.com/solana-labs/solana-pay/issues/100

let me know your thoughts, thanks.

Ajayneueworld:
> @Ajayneueworld please ask developer support questions on Discord. We don't use Github for this, especially not on unrelated pull requests.

vpontis:
I think we shouldn't use `HEAD` as described in the new spec:

- `HEAD` is a non standard API request that is not supported by many API clients and frameworks
- `HEAD` is useful for browsers since it's built into the browser spec but it's not commonly used for APIs outside of html

Instead, we can define a `GET` request:

- the path of the request will default to the same `link` parameter 
- the route will return a JSON object rather than requiring the requester to inspect the Headers

SPEC.md
```diff
@@ -8,80 +8,243 @@ Rough consensus on this spec has been reached, and implementations exist in Phan
...
+
+For example, this might be the name of an item being purchased, a discount applied to the purchase, or a thank you note. The wallet should display the value to the user.
+
+The wallet and application should allow additional fields in the request body and response body, which may be added by future specification.
+
+### Example
+
+##### URL describing a transaction request.
+```
+solana:https://mvines.com/solana-pay
```
antonok-edm, line 192: I'd recommend `example.com` for the URLs, rather than a parked domain that's currently available for purchase.

jordansexton: Yeah, that's probably a good idea.

jordansexton: Fixed!

PR #80, opened 2022-02-13
Fix the error in example 

Anmolnoor:
Fix the error in the example file index.ts in lines number 55 and 90 

💥💥💥 error TS2345: Argument of type 'BigNumber | undefined' is not assignable to parameter of type 'BigNumber'. Type 'undefined' is not assignable to type 'BigNumber'. 💥💥💥

Screenshot from 2022-02-13 09-10-05

😀 after adding the amount as BigNumber it works 😀
Screenshot from 2022-02-13 09-11-06

### DIFF
--- a/core/example/index.ts
+++ b/core/example/index.ts
@@ -53,7 +53,7 @@ import { createTransaction, encodeURL, findTransactionSignature, parseURL, valid
     }
 
     // Create a transaction to transfer native SOL or SPL tokens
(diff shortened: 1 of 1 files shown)


jordansexton:
You don't need to cast here. What you're looking for is a [non-null assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#non-null-assertion-operator-postfix-).

jordansexton:
Thanks!

PR #85, opened 2022-02-16
fixed broken links in docs/src/core

gautam2002:
The image relative links were not working inside docs/src/core/  merchant_integration.MD and overview.md
I created new folder images inside docs/src/images and moved necessary images from docs/static/img  and pointed the relative path of images to the above-broken readme


jordansexton:
Thanks for this fix! Can you revert all the code formatting changes though? We used Prettier to format the code.

gautam2002:
Sorry, can you guide me on or elaborate more on reverting the code formatting. But I also used Prettier.

PR #89, opened 2022-02-19
Set amount value if amount field is present in query string

Pancrisp:
I've noticed the amount isn't set even when an `amount` key is specified in the query string, requiring the user to re-enter an amount on screen.

This fixes that.


jordansexton:
Sorry, this isn't a bug or a design goal for now. The point of sale app is a proof of concept of using Solana Pay. Developers are encouraged to fork, modify, and deploy their own versions.

PR #95, opened 2022-03-03
Migrate the point-of-sale app to be fully NextJS

mcintyre94:
- Before this PR the point-of-sale app is split between `/client` (built etc with parcel, client-side rendered) and `/server` (the NextJS API to support transaction requests)
- This PR moves all the client code into the NextJS API, and flattens the directory so it's all under `/point-of-sale` with no `/client` or `/server` sub-directories
- The benefit of this is that users need to only deploy one NextJS app, instead of separately deploying the /client and /server

Demo deployed to Vercel: https://solana-pay-nextjs.vercel.app?recipient=...&label=...
eg. https://solana-pay-nextjs.vercel.app/?label=Solana+Pay&recipient=GvHeR432g7MjN9uKyX3Dzg66TqwrEWgANLnnFZXMeyyj

To run locally, use `yarn dev` in one terminal tab, `yarn proxy` in another, and visit https://localhost:3001/

The demo deployment uses an NPM published version of `/core` but is otherwise identical to this branch ([diff here](https://github.com/mcintyre94/solana-pay/compare/cm/link-request-nextjs...cm/link-request-nextjs-deploy))

To the best of my knowledge, this app is feature-equivalent with the existing one: https://solana-labs.github.io/solana-pay/app/new?recipient=GvHeR432g7MjN9uKyX3Dzg66TqwrEWgANLnnFZXMeyyj&label=Solana+Pay

Notes:

- I've minimised next-specific code as much as I can, but the following are areas where it wouldn't make sense to avoid it:
  - code that previously used react-router now uses Next equivalents (eg. `useRouter()` or `<Link>`)
  - The redirect from `/` to `/new` is done using a Next rewrite
  - The `recipient` and `label` query params are handled in the `_app.tsx` `getInitialProps` function, which makes SSR work smoothly (otherwise we'd SSR the logo when those params are there)
- I needed to add `postcss-nesting` because Next default config doesn't support nested CSS. With this all CSS is copied unchanged (although files are renamed to x.module.css for Next). The rest of postcss.config.json is [NextJS default](https://nextjs.org/docs/advanced-features/customizing-postcss-config#customizing-plugins). The next import format for CSS modules is also slightly different.
- I've added `tsc-esm` as a dependency to `/core` and use `npx` to invoke it. Not strictly necessary because I'm not building core in vercel, but these changes will make it work next time someone does! 
- I've bumped `@solana/spl-token` to 0.2.0 and `@solana/web3.js` to its newest version, in both core and point-of-sale
- I've changed some `useLayoutEffect` to `useEffect` to avoid an SSR warning. AFAICT there's no difference in these cases, but I can change them back if required
- `baseUrl` is new config, required to build URLs with SSR. I'm just using the `host` header for this and it should just work as long as https is used (which is required for Solana transaction requests anyway)
- I've only copied from `/client` what was needed to render each page/component etc. There are some things (like the `.svg` images directory) which are deleted in this PR and not copied because they don't seem to be needed


mcintyre94:
Found a bug in the deployed app: the QR code link has localhost:3001 embedded. Will look into it!

Using a QR code reader I'm seeing eg.: `solana:https%3A%2F%2Flocalhost%3A3001%2Fapi%3Frecipient%3DGvHeR432g7MjN9uKyX3Dzg66TqwrEWgANLnnFZXMeyyj%26amount%3D0.1%26reference%3D3FXF17E9jkuVfMmE8zk2j7vHqHck1deyUhYFh8XJbsrQ?label=Solana+Pay`

jordansexton:
Fantastic work on this, thank you! I refactored back into having client/server directories within the Next.js src just because I want the separation of concerns to be really clear to people building off this proof of concept. I also restored the images -- these are the original source files, so while they aren't used, I like to keep them around in case we make changes.

PR #103, opened 2022-03-12
Update Next app: handle GET request per transaction request spec, add optional message

mcintyre94:
This PR updates the point-of-sale next app to match the latest version of the spec as fully as possible:

- Implement the GET request described by the spec. This returns `{label, icon}`
  - Label is the URL param, which is now part of the encoded params in the QR code
  - Icon is the Solana Pay logo, which is now in `public/` and therefore available at `https://$HOST/SolanaPayLogo.svg`, eg. https://solana-pay-nextjs-git-cm-next-api-get-deploy-mcintyre1994.vercel.app/api?recipient=GvHeR432g7MjN9uKyX3Dzg66TqwrEWgANLnnFZXMeyyj&amount=1&reference=FrmaUM4eZ5RQhPMUdGWp1549aXUpttkFbEhmKYS6ZcJa&label=Solana%20Pay&message=thanks%21

- Add the optional `message` to the `POST` response.
  - `message` is an optional query param. If present it is part of the encoded params in the QR code
  - If present, it's returned alongside the transaction: `{"transaction": "AQA...", "message": "Thanks for your purchase!"}
eg.
```
$ curl --location --request POST 'https://solana-pay-nextjs-git-cm-next-api-get-deploy-mcintyre1994.vercel.app/api?recipient=GvHeR432g7MjN9uKyX3Dzg66TqwrEWgANLnnFZXMeyyj&amount=1&reference=FrmaUM4eZ5RQhPMUdGWp1549aXUpttkFbEhmKYS6ZcJa&label=Solana%20Pay&message=thanks%21' \
--header 'Content-Type: application/json' \
--data-raw '{
    "account": "Fkc4FN7PPhyGsAcHPW3dBBJ4BvtYkDr2rBFBgFpvy3nB"
}'
{"transaction":"AQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAABAAIE2y86zXzeWGYwGouJt6FEbb8nLg5UPMEz6t2CJWVD0PTshehNSVaLseABN3rUz0lfWemVAnXkDZDF5aiP6N3gggAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA3MN654JwNzCcGBkVUt92B5kq6JK7Z8eVbENh6hZIsZfIZfX07NBlEmoa1a2HhIUCV6C8XG5W2Rn3Iud15xfc3AECAwABAwwCAAAAAMqaOwAAAAA=","message":"thanks!"}
```

- Fix `getServerSideProps` which needs to be in the top-level `pages/` component (was in the `client/pages` one). Also copied that into the `NewPage` because that can be client rendered (click cancel on pending)

@jordansexton


point-of-sale/src/client/hooks/usePayment.ts
```diff
@@ -16,7 +16,6 @@ export interface PaymentContextState {
     amount: BigNumber | undefined;
     setAmount(amount: BigNumber | undefined): void;
     message: string | undefined;
```
jordansexton, line 18: Do we need this as part of the Payment context since it's already on the Config context?

point-of-sale/src/pages/new.tsx
```diff
@@ -1 +1,12 @@
 export { default } from '../client/components/pages/NewPage';
+
+export function getServerSideProps() {
```
jordansexton, line 3: Hmm, do you think it's better to export these here? My inclination is to place them with the component, but I don't have a strong opinion (or know if it makes a difference).

point-of-sale/src/pages/new.tsx
```diff
@@ -1 +1,12 @@
...
+  console.log('Running PendingPage getServerSideProps');
```
jordansexton, line 4: nit: `PendingPage` -> `NewPage`

Might not need the console log at all though.

point-of-sale/src/server/api/index.ts
```diff
@@ -59,7 +84,24 @@ const index: NextApiHandler<{ transaction: string }> = async (request, response)
     });
     const base64 = serialized.toString('base64');
 
-    response.status(200).send({ transaction: base64 });
+    response.status(200).send({ transaction: base64, message });
+}
+
+const index: NextApiHandler<GetResponse | PostResponse> = async (request, response) => {
```
jordansexton, line 90: Really nice explicit structure here.

jordansexton (commented):
Awesome, thanks for tackling this so quickly!

point-of-sale/src/client/hooks/usePayment.ts
mcintyre94: Oops got a bit mixed up there. I've removed it from the payment context, it's now just consumed by `PaymentProvider` to make the URL

point-of-sale/src/pages/new.tsx
mcintyre94: Agreed, didn't realise exporting from the component would work. Updated to `export { default , getServerSideProps } from '../client/components/pages/NewPage';` which sorts it and I think that's the better pattern

jordansexton:
Fantastic work, thanks so much for handling this.

PR #114, opened 2022-03-31
Update README.md

flodef:
Change the app opening adress as it was displaying solana pay logo, instead of the real app.

### DIFF
--- a/point-of-sale/README.md
+++ b/point-of-sale/README.md
@@ -63,7 +63,7 @@ yarn start
 
 ### Open the point of sale app
 ```shell
-open "http://localhost:1234?recipient=Your+Merchant+Address&label=Your+Store+Name"
+open "http://localhost:1234/new?recipient=Your+Merchant+Address&label=Your+Store+Name"
 ```
 
 ## Accepting USDC on Mainnet


jordansexton:
This is by design. It will redirect to `/new` if valid query params are provided. Check our label and recipient query params to make sure they are correct.

flodef:
You're right, just tested, it worked !
I find it a bit confusing anyway. Thanks for the quick answer.
And keep up the good work.

jordansexton:
Here's where we check for the params: https://github.com/solana-labs/solana-pay/blob/fb40bfc0a3a43824a4657b3d2a0ba67d3867c09f/point-of-sale/src/components/routes/RootRoute.tsx#L26-L42

Here's where we do the redirect: https://github.com/solana-labs/solana-pay/blob/fb40bfc0a3a43824a4657b3d2a0ba67d3867c09f/point-of-sale/src/components/routes/RootRoute.tsx#L26-L42

The redirect will only occur if the params are provided, and especially if the recipient is a valid pubkey.

flodef:
I see. That sure helps a lot !! Thank you :-D

PR #115, opened 2022-04-05
Support wallet disambiguation and deeplinking in protocol spec, url encoding, and url parsing

Saqif280:
# Problem Overview

This branch addresses requests in https://github.com/solana-labs/solana-pay/issues/100 to update the Solana Pay protocol specification to support wallet disambiguation in iOS clients and deeplinking to address some security vulnerabilities.

# Solution & Decision Log
The open issue proposes several changes; this PR is the first step towards implementing those changes from Solana Pay's end. In addition to this, there will need to be followup PRs to fully close the issue.

## Handling the Spec Changes
The issue poster suggested to allow for a url like `https://phantom.app/code?` instead of the current url scheme `solana:`. However, the suggested URL has a path, which should be used for the `recipient` according to the current protocol spec. There are a few options of how we might want to handle this:
1. We can make a breaking change to the protocol spec and require `recipient` as a query parameter. I'm avoiding this since breaking changes are costly.
2. We can continue to keep the `recipient` in the path, and just parse it out of the end of the wallet URL (in the cases where we use a wallet URL). This is a bit of a rigid solution since it relies on wallet applications registering a wildcard route to handle `recipient` in the path. The parsing also feels a little hacky to me, so I’m not a fan of this one.
3. We can allow for a `recipient` query parameter. This will be encoded in the URL when a wallet is provided, otherwise it will default to putting `recipient` in the path. This solution is backwards compatible and flexible so I decided to implement this.

## Supporting Wallet Disambiguation
This solution creates a TypeScript enum for tracking supported wallets called `SupportedWallet`. Wallets can create PRs to add their wallet and their registered url to this enum. This branch leaves the enum empty for now, which means we can't build much around this until enum entries are added.

# Testing
- Documentation changes were manually verified in local docusaurus
- Code changes were tested via unit tests
  - Note: `encodeURL.test.ts` will need additional tests once supported wallets have been added to the `SupportedWallet` enum.

# Next Steps
1. System diagrams in documentation need to be updated to reflect complete call paths (given the new `request` query param usage). _This could be included in this PR but I don't have access to the files for these diagrams._
2. Documentation needs to be added on how to set up authentication servers for merchants and how to interact with authentication servers as wallets. _This could also be included in this PR, but would be confusing without the updated system diagrams._
3. The [Point of Sale](https://github.com/solana-labs/solana-pay/tree/master/point-of-sale) package needs to be updated to provide UIs for wallet selection.

### DIFF
--- a/core/example/payment-flow-merchant/main.ts
+++ b/core/example/payment-flow-merchant/main.ts
@@ -105,14 +105,7 @@ async function main() {
     console.log('\n6. 🔗 Validate transaction \n');
 
     try {
-        await validateTransactionSignature(
-            connection,
-            signature,
-            MERCHANT_WALLET,
-            amount,
-            undefined,
-            reference
-        );
+        await validateTransactionSignature(connection, signature, MERCHANT_WALLET, amount, undefined, reference);
 
         // Update payment status
         paymentStatus = 'validated';
--- a/SPEC.md
+++ b/SPEC.md
@@ -18,11 +18,15 @@ By standardizing a simple approach to solving those problems, we ensure compatib
 
 ## Specification
 ```
-solana:<recipient>?amount=<amount>&label=<label>&message=<message>&memo=<memo>&reference=<reference>
+https://some-solana-wallet.com/?recipient=<recipient>&amount=<amount>&label=<label>&message=<message>&memo=<memo>&reference=<reference>&request=<request>
+```
+or
+```
+solana:<recipient>?amount=<amount>&label=<label>&message=<message>&memo=<memo>&reference=<reference>&request=<request>
 ```
 
 ### Recipient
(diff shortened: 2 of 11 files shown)


core/example/payment-flow-merchant/main.ts
```diff
@@ -105,14 +105,7 @@ async function main() {
...
     try {
-        await validateTransactionSignature(
-            connection,
-            signature,
-            MERCHANT_WALLET,
-            amount,
-            undefined,
-            reference
-        );
+        await validateTransactionSignature(connection, signature, MERCHANT_WALLET, amount, undefined, reference);
```
Saqif280, line 108: I didn't touch this file, I think my prettier caught it but going to leave in this PR

jordansexton:
This is a breaking change, and is outdated, so probably won't be included.

PR #117, opened 2022-04-13
transaction request docs

cogoo:
WIP: Docs for transaction requests.

PR includes changes from https://github.com/solana-labs/solana-pay/pull/77

### DIFF
--- a/docs/src/core/transaction-request/MERCHANT_INTEGRATION.md
+++ b/docs/src/core/transaction-request/MERCHANT_INTEGRATION.md
@@ -199,27 +199,6 @@ async function createSplTransferIx(sender, connection) {
 
 For our example, we create a simple transfer for a SPL token, serialize the transaction, and base64 encode it.
 
-<!--
-### Signed vs Unsigned Transactions
-
-The `POST` endpoint may respond with a partially or fully signed transaction.
-
-If the transaction signatures are nonempty:
-
-- The application must set the feePayer to the public key of the first signature.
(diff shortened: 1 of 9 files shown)


docs/src/core/transaction-request/MERCHANT_INTEGRATION.md
```diff
@@ -0,0 +1,270 @@
...
+
+        // create the transaction
+        const transaction = new Transaction();
+
+        // add the instruction to the transaction
+        transaction.add(splTransferIx);
+
+        // Serialize and return the unsigned transaction.
+        const serializedTransaction = transaction.serialize({
+            verifySignatures: false,
```
mcintyre94, line 134: Nit: this doesn't need to be false in my experience. Probably worth keeping it true so that if the TX is partially signed then those signatures that are present are validated.

jordansexton: Hmm. Why would we need to verify our own signature?

mcintyre94: I don't think it's super important, but in theory I guess the server could be requesting other signatures by some channel before returning the transaction or there could be some issue with a signature, since there aren't any restrictions on what the transaction could be that's generated. It's definitely a nit, but IIRC the default is true and I don't think there's a case where making it false is necessary. If there was some issue with a signature of the returned transaction and we use false then it'll presumably error when the wallet tries to sign/send the transaction, so I think it'd generally be better to have it error here instead

PR #119, opened 2022-04-14
Allow ATA owner to be off the curve

bricestacey:
I have been working on an integration that allows a program to accept payment using Solana Pay.

When getting the payer and recipient ATA, Solana Pay SDK uses the default value for `allowOwnerOffCurve` which happens to be `false`. This makes it a bit harder if one wanted to integrate Solana Pay with a program, e.g. when a PDA is the owner of the recipient account. Is there any reason not to change the default behavior to allow Token owners off the curve?

I believe an existing workaround is to have an owner on the curve and use the Approve instruction and delegate the full amount to the PDA. It's a bit clumsy though.

### DIFF
--- a/core/src/createTransaction.ts
+++ b/core/src/createTransaction.ts
@@ -99,13 +99,13 @@ export async function createTransaction(
         amount = amount.times(TEN.pow(mint.decimals)).integerValue(BigNumber.ROUND_FLOOR);
 
         // Get the payer's ATA and check that the account exists and can send tokens
-        const payerATA = await getAssociatedTokenAddress(splToken, payer);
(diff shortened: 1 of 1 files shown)


jordansexton:
Sending tokens to programs isn't a design goal of Solana Pay transfer requests. More importantly, changing the `createTransaction` function in the SDK isn't going to change how wallets that support Solana Pay actually create transactions from the URL, so I'm afraid this PR won't do anything.

However, you can accomplish this (and any other kind of custom transaction) with transaction requests. See #77 and #117 which are going to be merged within a few days, and are already supported in Phantom, Solflare, and Glow (in beta/TestFlight pending app store review).

PR #123, opened 2022-04-23
Basic instructions for deploying POS app on Vercel

cogoo:
Basic instructions for deploying POS app on Vercel

Commits:
- docs(images): pOS deployment reference images
- docs(point of sale): vercel deployment instructions for point of sale app


cogoo:
Reminder:

Once core SDK is deployed, we have to update the version of `@solana/pay` in `point-of-sale/package.json`

point-of-sale/README.md
```diff
@@ -121,11 +121,50 @@ When you're done, it should look like this:
...
+Configure the environment variables:
+
+![](./setup/5.Environment_variables.png)
+
+> Make sure you select the cluster endpoint you want to connect to.
+
+```
+CLUSTER_ENDPOINT=https://api.devnet.solana.com
+RATE_LIMIT=10
+RATE_LIMIT_INTERVAL=60
```
jordansexton, line 161: We can skip these two env vars here I think. If we add a default endpoint in the code, we can skip all 3 and the entire env var config input step in the screenshots.

cogoo:
It's worth having the environment variables section there, since the values are actually configurable, but toned down the importance since they're optional

PR #138, opened 2022-07-09
Refactor context

tranminhwang:
Right now I'm seeing React Context code written in hook files, so I made this pull request to include them in Context related files for clarity

### DIFF
--- a/point-of-sale/src/client/components/contexts/ConfigProvider.tsx
+++ b/point-of-sale/src/client/components/contexts/ConfigProvider.tsx
@@ -1,10 +1,8 @@
 import { PublicKey } from '@solana/web3.js';
-import React, { FC, ReactElement, ReactNode } from 'react';
-import { ConfigContext } from '../../hooks/useConfig';
+import React, { FC, ReactElement, ReactNode, createContext } from 'react';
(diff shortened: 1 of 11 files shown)


jordansexton:
This is intentional. The context state interface is needed in the hooks without a reference to the implementation (the context provider). We have components (TSX files) consuming hooks (TS files), not the other way around.

Some other ways to refactor this are to

- Extract the context interface to its own file which the hook and the provider component both reference. This is kind of a drag because those files end up very anemic. A way around this could be up combine the interfaces of all contexts together but that's more conceptually confusing.
- Combine the state, hook, and provider all in one file. I tend not to do this because I like to have a clear separation of components and other stuff.

Thanks for thinking about these things! I'm going to close this because it's a bit hard to review and doesn't clearly improve on the maintainability.

tranminhwang:
hmmm, I still think we should separate them 🥲 cc @jordansexton

jordansexton:
What is gained from separating them? And which of the approaches do you think is best?

Important to remember that this point of sale app is just a proof of concept and that it's already largely served its purpose. It was never intended for production use or long term development.

PR #139, opened 2022-07-10
Remove redundant code

tranminhwang:
- [x] Remove the default value of connection because the variable CLUSTER_ENDPOINT already has a default value in the env file

### DIFF
--- a/point-of-sale/src/server/core/connection.ts
+++ b/point-of-sale/src/server/core/connection.ts
@@ -1,4 +1,4 @@
 import { Connection } from '@solana/web3.js';
 import { CLUSTER_ENDPOINT } from './env';
 
-export const connection = new Connection(CLUSTER_ENDPOINT || 'https://api.devnet.solana.com', 'confirmed');
+export const connection = new Connection(CLUSTER_ENDPOINT, 'confirmed');


point-of-sale/src/client/utils/arraysEqual.ts
```diff
@@ -1,5 +1,4 @@
 export function arraysEqual(a: any[], b: any[]) {
-    if (a === b) return true;
```
jordansexton, line 2: This probably shouldn't be removed, it's a short circuit if the same instance is passed in.

point-of-sale/src/server/core/connection.ts
```diff
@@ -1,4 +1,4 @@
 import { Connection } from '@solana/web3.js';
 import { CLUSTER_ENDPOINT } from './env';
 
-export const connection = new Connection(CLUSTER_ENDPOINT || 'https://api.devnet.solana.com', 'confirmed');
```
jordansexton, line 4: 👍

tranminhwang: @jordansexton  you can approve this merge quest 🥲

PR #141, opened 2022-07-27
FIX: typo in docs, Soalna -> Solana

soltoshi:
Fixes a typo I found when reading the docs for Solana Pay.


jordansexton:
Good catch, thanks!

PR #155, opened 2022-09-05
corrected amount decimalPlaces logic

Niten619:
While invoking **createTransfer** method for transferring native SOL and SPL tokens, the logic which checks whether the amount provided has greater precision than the SOL/mint or not (_line 119 & line 148 of core/src/createTransfer.ts_) will always throw **CreateTransferError('amount decimals invalid')** error even if we provide the amount value which is less than 9 decimals (eg. 0.5 or 1.00254) because the logic "**(amount.decimalPlaces() ?? 0 > SOL_DECIMALS)**" will always be True for any decimal values (whether it be less than 9 decimal (eg. 0.08) or greater than 9 decimal (eg. 0.0021547854). 
Adding parenthesis to the logic like so would solve this.


jordansexton:
Thanks for the description of the issue! Could you remove the core/app stuff from the PR though?

Niten619:
Copy that.. anything else sir?

jordansexton:
Thanks, much easier to review now. It looks like this fix is needed in a few more places though:

https://github.com/solana-labs/solana-pay/search?q=decimalPlaces&type=

flodef:
Thanks a lot, sir !

PR #159, opened 2022-09-14
Replaced BaseURL by window.location

flodef:
When using "navigate" after payment has been processed, the BaseURL is reset and it ends up to navigating to default location (which is set to "localhost" in App.ts).


flodef:
@jordansexton What about this PR ? Is there something to add or could it be commited ?

jordansexton:
With this change, do we need `baseURL` as a config anymore? Can we replace other uses of it the same way?

flodef:
> With this change, do we need `baseURL` as a config anymore? Can we replace other uses of it the same way?

Totally agree! Thanks a lot for this remark.
As previously, didn't get far enough into refactorisation.

PR #161, opened 2022-09-22
Check for amount in query params

Joshua-Ogbonna:
Hi, we were using the Solana-pay point-of-sale for a project at our workplace. We didn't want users to use the calculator as the amount would be coming from our product. This will enable users to pay the exact amount that is coming from the website. We were wondering if you could merge this PR. 

Basically, I added a condition in the pending page to check for amount params in the URL.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,153 @@
+{
+  "name": "solana-pay-1",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "dependencies": {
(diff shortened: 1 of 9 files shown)


jordansexton:
The point of sale app is only a reference implementation. It's not intended to accommodate for new use cases. Many of the changes in this PR are specific to your application, business logic, even your website URLs. This should probably be a PR against your own repo :)

Joshua-Ogbonna:
Okay. Thank you very much.

The application has helped me a lot. I later found an error with my
implementation.

PR #172, opened 2022-11-25
Update CreateTransfer : deprecated + refactoring

flodef:
Replaced deprecated method getRecentBlockhash by getLatestBlockhash  Code refactoring in checkAccount()

### DIFF
--- a/core/src/createTransfer.ts
+++ b/core/src/createTransfer.ts
@@ -106,7 +106,7 @@ async function createSystemInstruction(
 
     // Check that the recipient is valid native accounts
     const recipientInfo = await connection.getAccountInfo(recipient);
-    if (!recipientInfo) throw new CreateTransferError('sender not found');
+    if (!recipientInfo) throw new CreateTransferError('recipient not found');
     if (!recipientInfo.owner.equals(SystemProgram.programId)) throw new CreateTransferError('recipient owner invalid');
     if (recipientInfo.executable) throw new CreateTransferError('recipient executable');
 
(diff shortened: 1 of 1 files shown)


core/src/createTransfer.ts
```diff
@@ -96,11 +89,7 @@ async function createSystemInstruction(
     connection: Connection
 ): Promise<TransactionInstruction> {
     // Check that the sender and recipient accounts exist
-    const senderInfo = await connection.getAccountInfo(sender);
-    if (!senderInfo) throw new CreateTransferError('sender not found');
-
-    const recipientInfo = await connection.getAccountInfo(recipient);
-    if (!recipientInfo) throw new CreateTransferError('recipient not found');
+    const {senderInfo, recipientInfo} = await checkAccount(connection, sender, recipient);
```
jordansexton, line 92: Hmm, this refactoring adds abstraction and doesn't really save much code. In addition, here you use the return values, and in the other function, you discard them. This suggests the abstraction isn't a great fit.

Overall, what we probably _really_ want to refactor to is something that expresses this:

- For system transfers, check to make sure the sender exists†, and _maybe_ check to make sure the recipient exists (the argument against is that zero balance should probably be allowed).
- For token transfers, check to make sure the _sender's ATA_ exists, and also check to to make sure the _recipient's ATA_ exists. We don't actually care if the sender† or recipient _System_ accounts exist or have SOL.

† We might also want to have a check for the SOL balance of the sender, since the sender is the expected fee payer. Probably this would use `Connection#getFeeForMessage` rather than hardcoding 5000 lamports as the cost. For System transfers, this will need to be calculated in addition to the transfer amount.

jordansexton (changes requested):
Thanks for this! The `latestBlockhash` change is good, and refactoring the account checks is a good idea, but doesn't go far enough yet.

flodef:
@jordansexton Any ETA on commit ?

flodef:
The build failed because of the recent change in the folder structure (pos ==> examples/pos).
But the file in itself is building correctly :-D

flodef:
> Thanks for this! The `latestBlockhash` change is good, and refactoring the account checks is a good idea, but doesn't go far enough yet.

Excellent comment.
Thanks for all the explanation that were crystal clear.
It's done.

PR #173, opened 2022-11-28
Add an NFT create example

mcintyre94:
This is an example where we use transaction requests to airdrop a freshly created NFT to the buyer as part of the transaction.

See new readme for details of the new example: https://github.com/solana-labs/solana-pay/tree/nft-example/examples/nft-create#readme

- Moved `point-of-sale` into a new `examples/` directory
- Add new `nft-create` example


jordansexton (approved):
LGTM!

PR #176, opened 2022-12-11
BackButton style proposal

flodef:
BackButton should be hidden when disabled + take less space for mobile phone with small resolution

### DIFF
--- a/examples/point-of-sale/src/client/components/buttons/BackButton.module.css
+++ b/examples/point-of-sale/src/client/components/buttons/BackButton.module.css
@@ -1,7 +1,7 @@
 .button {
-    height: 40px;
-    padding: 16px;
-    margin: 16px;
+    height: 48px;
+    padding: 12px;
+    margin: 12px;
     border: 0;
     display: inline-flex;
     align-items: center;
--- a/examples/point-of-sale/src/client/components/buttons/BackButton.module.css
+++ b/examples/point-of-sale/src/client/components/buttons/BackButton.module.css
@@ -1,5 +1,5 @@
 .button {
-    height: 48px;
+    height: 40px;
     padding: 16px;
     margin: 16px;
     border: 0;
@@ -9,4 +9,11 @@
     font-weight: 500;
     background: none;
     cursor: pointer;
+    visibility: visible;
+}
+
+.button[disabled] {
+    cursor: default;
+    opacity: 0.5;
+    visibility: hidden;
 }


examples/point-of-sale/src/client/components/buttons/BackButton.module.css
```diff
@@ -1,5 +1,5 @@
 .button {
-    height: 48px;
+    height: 40px;
```
steveluscher, line 2: 48px is actually the Apple Human Interface Guideline for the size of tap targets. You can make the size of the icon smaller, but keep the tap target 48px.

steveluscher (changes requested):
Just a drive by usability blocker; I’ll let someone else comment on whether the design proposal is mergeable in general.

examples/point-of-sale/src/client/components/buttons/BackButton.module.css
steveluscher: https://web.dev/accessible-tap-targets/

flodef: Thanks for the teaching!
I'll pay attention to that from now on...
Testing with a small device like IPhone SE to handle the case where the Generate button is not visible anymore.
Changing the margin / padding does not change the layout but keep the button size :-)

jordansexton:
This app is meant as an example and proof of concept only. I'm hesitant to change the styles here, and time is limited to test these changes.

steveluscher:
Hey @flodef. The code in `examples/point-of-sale` is what's called a ‘reference implementation.’ Reference implementations are demonstrations of a specification – in this case Solana Pay – in action. They're generally not meant to be deployed in production as-is, but rather to act as a _reference_ for others.

Changes to the reference implementation that relate to its correctness or adherence to the Solana Pay specification are in scope, and changes that affect the aesthetics or usability of the example app itself are out of scope.

You should for sure fork the reference implementation and open source something targeted toward the exact user you've described though! You could be a huge part of onboarding that next billion through the creation of something free, fast, and delightful to use. We created and open sourced the Solana Pay spec to encourage exactly that.

Thank you for your contributions 🚀

flodef:
Always happy to read from you. Thanks a lot for the precisions. 
As I strongly suggested to jordan, it could be wise to precise the scope of PR and the purpose of it directly in the README.md from "examples/point-of-sale" (it went from "point-of-sale" to "examples/point-of-sale" very recently).

Like you saw, I've already forked the reference implementation and it's live. 
Sharing it from solana-labs' repository for dev to start building from a prod version instead of reference implementation could save other devs a lot of time. It's working out of the box, and you just have to change settings in vercel. You don't even have to be a dev to use / customise it.

But, in the current state, that's not gonna happen. No Solana dev will come to flodef's github repository to start a Point of Sale based on Solana Pay. Like me, they'll come on solana-labs', find the reference implementation and start building from it, wasting time, instead of improving each other's work. 
In the end, every dev is building is own point-of-sale implementation in its small corner...

PR #179, opened 2022-12-12
Num pad

flodef:
Set constant for decimals and maxValue that can be entered.
Avoid overflow (like typing 55555555555555555555555555).


jordansexton:
Thanks, but this kind of customization can be performed by developers. The point of sale app is just an example and proof of concept.

flodef:
Screenshot 2022-12-13 00 46 03

I would definitely not be proud to let that happen on one of my app (proof of concept or not) !

PR #180, opened 2022-12-12
End point

flodef:
Make the CLUSTER_ENDPOINT available to browser to be able to change it from the environment settings all at once.
Add constants to be able to select the currency and infers its dependencies directly in MainApp : all the hard coded variables are now dynamic (extracted from constants)


jordansexton:
Thanks, but we aren't making many changes (e.g. currency selection) to the Point of Sale example.

flodef:
> Thanks, but we aren't making many changes (e.g. currency selection) to the Point of Sale example.

What ?!
You prefer to have hard coded values and ask people to change the currency by writing code (see how in the README.md file), instead of just changing a setting parameter in Vercel ???

PR #181, opened 2022-12-12
Package

flodef:
Upgrade packages to latest versions : this allow to use new mobile wallet adapter, and fixes some bugs.


examples/point-of-sale/package.json
```diff
@@ -6,7 +6,7 @@
     "repository": "https://github.com/solana-labs/solana-pay",
     "license": "Apache-2.0",
     "engines": {
-        "node": ">=16"
+        "node": "18"
```
jordansexton, line 9: Please don't change the Node version.

examples/point-of-sale/package.json
```diff
@@ -20,47 +20,46 @@
...
-        "bs58": "^4.0.1",
-        "cache-manager": "^4.1.0",
+        "@solana/spl-token": "^0.3.6",
+        "@solana/wallet-adapter-base": "^0.9.19",
+        "@solana/wallet-adapter-react": "^0.15.25",
+        "@solana/wallet-adapter-react-ui": "^0.9.23",
+        "@solana/wallet-adapter-wallets": "^0.19.7",
+        "@solana/web3.js": "^1.69.0",
+        "bignumber.js": "^9.1.1",
+        "bs58": "^5.0.0",
```
jordansexton, line 30: I don't think this should be changed, web3.js still uses v4 AFAIK.

examples/point-of-sale/package.json
```diff
@@ -20,47 +20,46 @@
...
+        "cache-manager": "^5.1.4",
```
jordansexton, line 31: What breaking changes does this introduce?

examples/point-of-sale/package.json
```diff
@@ -20,47 +20,46 @@
...
         "react-circular-progressbar": "^2.1.0",
         "react-dom": "^18.2.0",
-        "react-responsive": "^9.0.0-beta.10"
+        "react-responsive": "^9.0.2"
     },
     "devDependencies": {
         "@types/bs58": "^4.0.1",
-        "@types/cache-manager": "^4.0.1",
-        "@types/cors": "^2.8.12",
+        "@types/cache-manager": "^4.0.2",
```
jordansexton, line 44: The types here are v4, while the package above is v5. They should probably match! If the types haven't been updated for v5, we shouldn't update.

examples/point-of-sale/package.json
```diff
@@ -20,47 +20,46 @@
...
+        "@types/cors": "^2.8.13",
         "@types/eslint-plugin-prettier": "^3.1.0",
-        "@types/express": "^4.17.13",
-        "@types/express-rate-limit": "^6.0.0",
```
jordansexton, line 48: Why was this removed?

examples/point-of-sale/package.json
```diff
@@ -20,47 +20,46 @@
...
+        "@types/prettier": "^2.7.1",
+        "@types/react": "^18.0.26",
+        "@types/react-dom": "^18.0.9",
+        "eslint": "^8.29.0",
+        "eslint-config-next": "^13.0.6",
         "eslint-config-prettier": "^8.5.0",
         "eslint-plugin-prettier": "^4.2.1",
-        "local-ssl-proxy": "^1.3.0",
-        "next": "^12.2.5",
+        "local-ssl-proxy": "^1.2.3",
```
jordansexton, line 56: Why was this downgraded?

examples/point-of-sale/src/server/core/cache.ts
```diff
@@ -1,5 +1,3 @@
 import cacheManager from 'cache-manager';
-// @ts-ignore (TS7016) There is no type definition for this at DefinitelyTyped.
-import MemoryStore from 'cache-manager/lib/stores/memory';
 
-export const cache = cacheManager.caching({ store: MemoryStore, max: 1000, ttl: 120 /*seconds*/ });
+export const cache = cacheManager.caching('memory', { max: 1000, ttl: 120 /*seconds*/ });
```
jordansexton, line 3: Is this change the only reason to update the package?

jordansexton (changes requested):
Thanks for the PR. I think this should be more limited in scope. We probably don't need to update every package!

examples/point-of-sale/package.json
flodef: Either you select "16", or you select "18", but setting ">=16" trigger a warning when building. Using vercel, the app build with whatever node version is in the setting. If you don't like "18", set "16" insead, please.

examples/point-of-sale/package.json
flodef: bs58 v5 was release 10 months ago !!
I tested it and it's working fine with web3.js.

examples/point-of-sale/package.json
flodef: What do you mean ? 
Is it breaking something is the app ? No.
Is it worth the risk to update those packages ? Yes, removed a lot of warning while compiling.

examples/point-of-sale/package.json
flodef: Have to agree on this one!

examples/point-of-sale/package.json
flodef: Again, generating warning during the build. Those are not needed.

examples/point-of-sale/package.json
flodef: Can't remember why, but agree on this one too!
Built and working great with last version.

examples/point-of-sale/src/server/core/cache.ts
flodef: Nope. 
This is a collateral.
Main reasons are to use solana mobile adapter and get rid of all the warning when compiling

PR #186, opened 2023-01-23
Add Gradient to QR code options

flodef:
Allows to put some color with Gradient in your QR code.

### DIFF
--- a/core/src/createQR.ts
+++ b/core/src/createQR.ts
@@ -7,6 +7,7 @@ import type {
     Mode,
     Options,
     TypeNumber,
+    Gradient,
 } from '@solana/qr-code-styling';
 import QRCodeStyling from '@solana/qr-code-styling';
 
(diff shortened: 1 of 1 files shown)


core/src/createQR.ts
```diff
@@ -22,8 +23,7 @@ export function createQR(url: string | URL, size = 512, background = 'white', co
     return new QRCodeStyling(createQROptions(url, size, background, color));
 }
 
-/** @ignore */
-export function createQROptions(url: string | URL, size = 512, background = 'white', color = 'black'): Options {
+function createQROptions(url: string | URL, size = 512, background = 'white', color = 'black', gradient?: Gradient): Options {
```
mcintyre94, line 26: Thanks for the PR! Can you keep this function exported please, we don't want to break existing callers

mcintyre94 (changes requested):
If we do this then I think we should also pass `gradient` to `createQR` to keep it mirroring the `createQROptions` params

mcintyre94:
Thanks again @flodef but we're not going to merge this. The underlying library `@solana/qr-code-styling` has a lot of options, and it's intentionally very easy to create your own `Options` object and create a QR code using any of these options. We're not going to expand this API which is only intended to provide the basic options and try to push people toward easily scanned QR codes by default.

core/src/createQR.ts
flodef: It's an optional parameter, so existing callers won't see any differences.
I tested it prior commiting and it works as before without the gradient param.

flodef:
Adding an **optional** Gradient option doesn't complicate anything but only add one more option to power users who need more, while keeping it as simple as before for previous users. So creating a QR code is still as simple as it was !
I don't get your point, but that's OK, if you took the decision to not merge, I'll kindly comply.
PS: I updated the file with your suggestion about adding it to `createQROptions` :-)

PR #197, opened 2023-02-28
Sdk updates  for proposed 1.1 spec

mcintyre94:
Marking as draft because we'll want to coordinate this release so we don't confuse application developers. Maybe a 1.1 or alpha branch?

This PR adds the spec changes to support the proposed/alpha 1.1 spec

- Redirect is added as an optional parameter on transfer request URLs. Note that for transaction requests redirects are part of the POST response which is out of scope for the SDK.
- A new `MessageSignRequestURLFields` is added with just a `link` field. Internally `encodeURL` encodes it in the same way as a transaction request

Notes:

- No SDK changes are made for transaction request error handling, because again they're just part of the POST response
- Should we deprecate the `label` and `message` fields of `TransactionRequestURLFields`? They're not in the spec and I haven't included them in the fields for message signing requests. I think they're superseded by `label` in the GET response, and `message` in the POST response. If we eventually remove them we could simplify the encoding since transaction request and message signing request would be the same structure. 
- I think using `encodeTransactionOrMessageSignRequestURL` is kinda hacky, but we can't differentiate `TransactionRequestURLFields` from `MessageSignRequestURLFields` at a typescript level

### DIFF
--- a/core/src/encodeURL.ts
+++ b/core/src/encodeURL.ts
@@ -1,5 +1,5 @@
 import { SOLANA_PROTOCOL } from './constants.js';
-import type { Amount, Label, Memo, Message, Recipient, References, SPLToken } from './types.js';
+import type { Amount, Label, Memo, Message, Recipient, References, SPLToken, Redirect } from './types.js';
 
 /**
  * Fields of a Solana Pay transaction request URL.
@@ -31,6 +31,8 @@ export interface TransferRequestURLFields {
     message?: Message;
     /** `memo` in the [Solana Pay spec](https://github.com/solana-labs/solana-pay/blob/master/SPEC.md#memo). */
     memo?: Memo;
+    /** `redirect` in the [Solana Pay spec](https://github.com/solana-labs/solana-pay/blob/master/SPEC1.1.md#redirect). */
+    redirect?: Redirect;
 }
 
 /**
@@ -68,6 +70,7 @@ function encodeTransferRequestURL({
     label,
     message,
     memo,
+    redirect,
 }: TransferRequestURLFields): URL {
     const pathname = recipient.toBase58();
     const url = new URL(SOLANA_PROTOCOL + pathname);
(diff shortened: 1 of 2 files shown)


core/src/encodeURL.ts
```diff
@@ -31,18 +31,34 @@ export interface TransferRequestURLFields {
...
     /** `memo` in the [Solana Pay spec](https://github.com/solana-labs/solana-pay/blob/master/SPEC.md#memo). */
     memo?: Memo;
+    /** `redirect` in the [Solana Pay spec](https://github.com/solana-labs/solana-pay/blob/master/SPEC1.1.md#redirect). */
+    redirect?: Redirect;
+}
+
+/**
+ * Fields of a Solana Pay message sign request URL.
+ */
+export interface MessageSignRequestURLFields {
```
jordansexton, line 41: I think this type can be (structurally) the same as `TransactionRequestURLFields`, right? Label and message can still be used.

core/src/encodeURL.ts
```diff
@@ -31,18 +31,34 @@ export interface TransferRequestURLFields {
...
+): URL {
+    return 'link' in fields ? encodeTransactionOrMessageSignRequestURL(fields) : encodeTransferRequestURL(fields);
 }
 
-function encodeTransactionRequestURL({ link, label, message }: TransactionRequestURLFields): URL {
+function encodeTransactionOrMessageSignRequestURL({
+    link,
+    label,
+    message,
+}: TransactionRequestURLFields | (MessageSignRequestURLFields & { label: undefined; message: undefined })): URL {
```
jordansexton, line 61: This can also be simplified if we allow those fields.

core/src/encodeURL.ts
jordansexton: I think maybe we should rename the type of request to a `Sign Message Request` in the spec and here.

jordansexton (commented):
Looks good! I'm going to do another round of review on the spec itself.

core/src/encodeURL.ts
mcintyre94: They can still be used, but it seems a bit confusing because wallets would presumably ignore them. Both fields are not inputs in the spec, and are also fields in server responses which I think makes it more confusing. We do already have this problem with transaction requests - I just wasn't sure whether it was a good idea to do the same here, and thought it might be better to move toward removing the fields from transaction request in the SDK instead.

PR #203, opened 2023-03-23
Read the amount from the URL if it is set

### LINKED ISSUE #202: Query issue: Amount does not show on pending page
http://localhost:3000/pending?recipient=54LnyoDbFvSXG4doBpfrZUoytQcJECWqkvDAjjh7sny7&amount=1&label=stark&message=great

The amount is set to 1 but, on the pending page it shows me 0. I have not used the new page to create an QR Code directly, I wanted to use it similar to other pay systems like paypal.

Any idea, what to change? I also saw the getServerSideProps, does this have to do anything with my issue?

`export function getServerSideProps() {
    // Required so getInitialProps re-runs on the server-side
    // If it runs on client-side then there's no req and the URL reading doesn't work
    // See https://nextjs.org/docs/api-reference/data-fetching/get-initial-props
    return {
        props: {},
    };
}`

Any help strongly appreciated. 🙌😎😎😎

mcintyre94:
Closes #202 

This issue is caused by the POS app using only local state for the `amount`, while other fields (like recipient and label) use the URL values. This PR sets the initial value of the `amount` state using the `amount` query param if it is present.

### DIFF
--- a/examples/point-of-sale/src/client/components/contexts/PaymentProvider.tsx
+++ b/examples/point-of-sale/src/client/components/contexts/PaymentProvider.tsx
@@ -11,6 +11,7 @@ import {
 import { useConnection, useWallet } from '@solana/wallet-adapter-react';
 import { ConfirmedSignatureInfo, Keypair, PublicKey, Transaction, TransactionSignature } from '@solana/web3.js';
 import BigNumber from 'bignumber.js';
+import { useRouter } from 'next/router';
 import React, { FC, ReactNode, useCallback, useEffect, useMemo, useState } from 'react';
 import { useConfig } from '../../hooks/useConfig';
 import { useNavigateWithQuery } from '../../hooks/useNavigateWithQuery';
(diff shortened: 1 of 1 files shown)


examples/point-of-sale/src/client/components/contexts/PaymentProvider.tsx
```diff
@@ -26,7 +27,16 @@ export const PaymentProvider: FC<PaymentProviderProps> = ({ children }) => {
...
 
-    const [amount, setAmount] = useState<BigNumber>();
+    const router = useRouter();
+
+    const [amount, setAmount] = useState<BigNumber | undefined>(() => {
+        // Use the amount query param as initial value if set
+        const { amount: queryAmount } = router.query;
+        if (!queryAmount) return undefined;
+        if (Array.isArray(queryAmount)) return new BigNumber(queryAmount[0]);
+        return new BigNumber(queryAmount)
```
jordansexton, line 37: Hmm, I think we might need to validate the amount in some way. If it doesn't pass a simple regex, we probably don't want to instantiate a `BigNumber` from it.

mcintyre94: I've added some validation (not NaN, finite, >0) using BigNumber functions. I think that's a bit more straightforward than adding our own regex, and since the amount is displayed on screen + in the wallet it's ok to allow whatever parsing flexibility BigNumber has here. Can add a regex instead if you'd prefer though, I don't have a strong view either way really.

jordansexton: Yeah, I think either is fine. My idea was simply to check it against `^\d+(\.\d+)?$` since this is (more or less) what describes a valid manual amount entry into the UI. Zero amounts are _technically_ probably okay to allow (they are allowed in the protocol) but I don't care enough to change this.

Issue #56, opened 2022-02-04
[todo] Add more robust retry logic to Point of Sale app


jordaaash:
The point of sale app should probably be more robust against RPC or network errors (either from Solana or the WiFi).

0xPratik:
Hey @jordansexton Can you please elaborate on it. I would like to work on this.

jordansexton:
Yeah, for sure. The goal is to increase the robustness of the app to connection issues.

We use a single RPC endpoint to [find transactions](https://github.com/solana-labs/solana-pay/blob/d6563f9c5d57ff13b981157a39100000a03c2ca6/point-of-sale/src/components/contexts/PaymentProvider.tsx#L113) in a loop until it's found. However, Phantom uses a different RPC by default (I think they use Triton) and then have some backup RPCs they use.

If our RPC node doesn't see the transaction that Phantom has sent yet, this causes latency to the user waiting for their transaction to confirm. It could be advantageous to poll several RPC nodes at once, from different RPC providers.

Because we use the same connection to [validate the transaction](https://github.com/solana-labs/solana-pay/blob/d6563f9c5d57ff13b981157a39100000a03c2ca6/point-of-sale/src/components/contexts/PaymentProvider.tsx#L143), we will want to reuse whichever connection the transaction was found using (since it's possible it was seen by one node and not another). It's also possible that we should try to validate the transaction over several connections too.

We would probably want all the connections to race to find and validate the transaction, so the merchant and customer sees their transaction confirmed as fast as possible.

Additionally, it would be nice to have something that pinged a known good to test the internet / WiFi connection from the PoS, so that we could indicate if there's an issue. This could be a separate PR, and should maybe be done before the more complicated RPC changes.

jordansexton:
After spiking this out, it may be worth considering if more robust connection management could be encapsulated as a small library, which we could then use here. I could see many apps having use for something like this.

nickbalestra:
Hey @jordansexton! 
I had a quick look into the codebase to see if I could help with this issue :) 
Set aside the small [PR](https://github.com/solana-labs/solana-pay/pull/133) I've reference as an idea on how we could to make sure we are online. Here some thoughts:

I see that the connection is set through the `ConnectionProvider` from '@solana/wallet-adapter-react'. 
And I guess that is what you mentioned with `We use a single RPC endpoint`. I like the idea of a small library that could handle multiple endpoints/connections! Take this just as an early sketch/idea: 

```typescript
[...]

PR #205, opened 2023-05-13
 Add more robust retry logic to Point of Sale app #56

lastemp:
This change adds two functions "findReferenceMultipleNodes" and "validateTransferMultipleNodes" from two new typescript files "validateTransferMultipleNodes.ts" and "findReferenceMultipleNodes.ts". These functions can be executed from Point of Sale app or any other application that interacts with Solana-pay.

The overall aim is to execute "findReference" and "validateTransfer" against multiple nodes to guarantee robustness. 

This solves the issue "Add more robust retry logic to Point of Sale app #56"

### DIFF
--- a/core/src/findReferenceMultipleNodes.ts
+++ b/core/src/findReferenceMultipleNodes.ts
@@ -1,0 +1,75 @@
+import { findReference } from "@solana/pay";
+import { ConfirmedSignatureInfo, Connection, Finality, PublicKey, SignaturesForAddressOptions } from '@solana/web3.js';
+import assert from 'assert';
+
+/**
+ * find Reference against multiple nodes to guarantee robustness.
+ * The endpoints are provided as string array.
+ * let endpoint1: string = 'https://api.devnet.solana.com';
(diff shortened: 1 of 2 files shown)


mcintyre94:
Hey, thanks for contributing this! Since #56 was opened there's been some work on a rewrite of web3js which allows using a custom transport at the connection layer. For example, Steven Luscher has tweeted an example of using this to round-robin requests to multiple RPC endpoints: https://twitter.com/steveluscher/status/1650794822687653888?s=20. This should allow us to keep behaviour like this outside of the Solana Pay library, and available to any code built on web3js. 

I don't want to add code to the core SDK that tries to address this - we want to keep this core unopinionated and flexible. The new web3js looks like the best solution so far. 

On the issue there was discussion of building a connection pooling library which could be used with this SDK or any other web3js code, and I think that's another potential solution. But I think this new web3js version is the most promising approach at the moment.

I'm going to close this for now, and will add a comment to #56 about this potential future solution. I know that this doesn't fix the issue in the immediate term, but we really don't want to add code that's quite this opinionated to the core SDK.

Thanks again for contributing, and let me know if you have any other questions! :)

PR #211, opened 2023-08-25
Add Ottr Wallet to supported list

lexer:


mcintyre94:
Thanks for this! We needed to update the list a bit so I've added Ottr and a few other wallets, and tidied this list up a bit in https://github.com/solana-labs/solana-pay/pull/212.

PR #220, opened 2023-12-11
Show VersionedTransaction in the tx requests doc

mcintyre94:
- I've only updated the docs here, ie the content at https://docs.solanapay.com/core/transaction-request/merchant-integration. We should update this repo to new web3js at some point, but I don't think it's worth rewriting any code for VersionedTransaction

- I'm doing `Buffer.from(serializedTransaction).toString('base64')` to convert the uint8array serialized transaction to a base64 string. This is pretty rubbish but goes away with the new web3js that has proper codecs for things like this. Not worth complicating the docs to do anything smarter here though.

### DIFF
--- a/docs/src/core/transaction-request/MERCHANT_INTEGRATION.md
+++ b/docs/src/core/transaction-request/MERCHANT_INTEGRATION.md
@@ -84,7 +84,7 @@ The `GET` endpoint should respond with two properties. `label` describes the sou
 The second part of the transaction request spec is the `POST` request.
 
 ```javascript
-import { clusterApiUrl, Connection, Keypair, PublicKey, Transaction } from '@solana/web3.js';
+import { clusterApiUrl, Connection, Keypair, PublicKey, VersionedTransaction } from '@solana/web3.js';
 import BigNumber from 'bignumber.js';
 import { createTransferCheckedInstruction, getAccount, getAssociatedTokenAddress, getMint } from '@solana/spl-token';
 import { TEN } from '@solana/pay';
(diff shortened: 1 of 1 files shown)


docs/src/core/transaction-request/MERCHANT_INTEGRATION.md
```diff
@@ -102,19 +102,21 @@ const post = async (request, response) => {
     // create spl transfer instruction
     const splTransferIx = await createSplTransferIx(sender, connection);
 
-    // create the transaction
-    const transaction = new Transaction();
-
-    // add the instruction to the transaction
-    transaction.add(splTransferIx);
+    const { blockhash } = await connection.getLatestBlockhash();
```
jordaaash, line 105: I might propose using the empty blockhash (`111...`) here as described in the spec. The wallet is supposed to fill this in if empty, which is better than providing one. Also getting a blockhash delays the response.

docs/src/core/transaction-request/MERCHANT_INTEGRATION.md
```diff
@@ -102,19 +102,21 @@ const post = async (request, response) => {
...
+            payerKey: sender,
+            recentBlockhash: blockhash,
+            // add the instruction to the transaction
+            instructions: [splTransferIx]
+        }).compileToV0Message()
+    )
+
+    const serializedTransaction = transaction.serialize()
+
+    const base64Transaction = Buffer.from(serializedTransaction).toString('base64');
```
jordaaash, line 119: We should be able to use `TextDecoder` here instead of `Buffer`, right?

jordaaash: Ignore me lol, TextDecoder/Encoder are for utf8 strings.

PR #244, opened 2024-10-13
fix(createTransfer): use  `getLatestBlockhash` instead  of deprecated…

melonges:
… `getRecentBlockhash`

this method [has been removed](https://solana.com/docs/rpc/deprecated/getrecentblockhash) from the new version of Solana RPC and is no longer available


gitteri:
Fixed by another PR

PR #272, opened 2025-10-18
impl: gill/kit solana pay

pkxro:
- gill compatible version of solana pay
- new package as to not break the existing web3js implementation
- 1:1 compatible with web3js solana pay
- includes same tests


gitteri (commented):
Can you add something to the readme about this addition? there are no docs for this right now.

PR #325, opened 2026-04-29
feat: add hmac request signing as auth method for pay start

MicaiahReid:


rust/.cargo/config.toml
```diff
@@ -1,5 +1,6 @@
 [alias]
 cli-install = "install --path crates/cli --locked --force"
+cli-install-gcp-kms = "install --path crates/cli --locked --force --features gcp_kms"
```
lgalabru: I don't think this worth being added here

rust/crates/cli/src/commands/server/start.rs
```diff
@@ -177,9 +194,7 @@ impl StartCommand {
         // Resolve config that doesn't need async.
         let currencies = resolve_operator_currencies(op, &self.currency);
 
-        let network = op
-            .and_then(|o| o.network.clone())
-            .unwrap_or_else(|| "mainnet".to_string());
+        let network = effective_server_network(op, sandbox);
```
lgalabru: Is this fixing a bug you're been experiencing?

MicaiahReid: Nah just an overeager codex

rust/.cargo/config.toml
MicaiahReid: Removed

rust/crates/cli/src/commands/server/start.rs
MicaiahReid: Removed

PR #341, opened 2026-05-04
ux: clarify Touch ID requirement for Keychain

### LINKED ISSUE #340: Clarify Touch ID requirement during macOS Keychain setup
## Summary

When running `pay setup` on macOS, the setup flow offers:

```text
macOS Keychain (Touch ID)
```

However, the current macOS Keychain backend requires Touch ID/biometrics to be available and configured. In environments where Touch ID is unavailable, setup can fail in a way that is confusing because it is not immediately clear that Touch ID is a hard requirement for using this backend with pay.

## User impact

A user can get stuck running `pay setup` on a Mac or macOS environment without Touch ID available. The prompt mentions Touch ID, but it may read like a description rather than a requirement.

Once the same flow is run on a machine with Touch ID enabled, setup works as expected.

## Expected behavior

The setup flow should make it clearer that Touch ID is required for the macOS Keychain backend.

For example:

```text
macOS Keychain (requires Touch ID)
```

If Touch ID authentication fails because it is unavailable or not configured, the error should include a short guidance message explaining that Touch ID must be available and configured on the Mac.

## Suggested fix

- Update the macOS backend picker label from:

```text
macOS Keychain (Touch ID)
```

to:

```text
macOS Keychain (requires Touch ID)
```

- Add guidance to non-cancel Touch ID authentication failures, preserving the original error and explaining that Touch ID is required to use macOS Keychain with pay.

This should be a UX-only change. It should not hide the Keychain option, and it should not change the underlying macOS authentication policy from biometrics to device-owner authentication.

foxytanuki:
## Summary

- Clarify that the macOS Keychain backend requires Touch ID during account setup.
- Add a short guidance message when Touch ID authentication fails for non-cancel backend errors.
- Keep the existing backend behavior unchanged.

Fixes #340.

## Background

A user reported getting stuck while running `pay setup` in a macOS environment where Touch ID was not available. The setup prompt currently says:

```text
macOS Keychain (Touch ID)
```

That mentions Touch ID, but it can read like a backend description rather than a hard requirement. Since pay's macOS Keychain backend uses Touch ID/biometrics for authentication, the setup experience is clearer if the requirement is explicit.

## Changes

The backend picker label now says:

```text
macOS Keychain (requires Touch ID)
```

For non-cancel Touch ID authentication failures, the original error is preserved and followed by guidance:

```text
Touch ID is required to use macOS Keychain with pay. Make sure Touch ID is available and configured on this Mac, then try again.
```

## Notes

This intentionally does not:

- hide the macOS Keychain option when Touch ID is unavailable
- change the Swift helper from `.deviceOwnerAuthenticationWithBiometrics` to `.deviceOwnerAuthentication`
- alter the underlying authentication/security behavior

This PR only improves the setup/authentication messaging.

## Validation

```bash
cargo fmt --check
cargo test -p pay setup
```

### DIFF
--- a/rust/crates/cli/src/commands/account/new.rs
+++ b/rust/crates/cli/src/commands/account/new.rs
@@ -291,7 +291,7 @@ pub fn pick_backend() -> pay_core::Result<String> {
     #[cfg(target_os = "macos")]
     let options = [Opt {
         id: "keychain",
-        label: "macOS Keychain (Touch ID)".into(),
+        label: "macOS Keychain (requires Touch ID)".into(),
     }];
 
     #[cfg(target_os = "linux")]
--- a/rust/crates/keystore/src/macos/mod.rs
+++ b/rust/crates/keystore/src/macos/mod.rs
@@ -38,7 +38,7 @@ impl AuthGate for TouchId {
             if is_user_cancel(&err) {
                 Err(Error::AuthDenied(err))
             } else {
-                Err(Error::Backend(err))
+                Err(Error::Backend(touch_id_unavailable_guidance(&err)))
             }
         }
     }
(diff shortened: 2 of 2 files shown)


lgalabru (approved):
Thanks @foxytanuki!

lgalabru (approved):

PR #342, opened 2026-05-04
docs: clarify Pay MCP selection guidance

Sarthib7:
## Summary

  Clarifies when agents should use Pay MCP versus free/local/web sources.

  This adds a lightweight pay-worthiness check whose goal is not to avoid cheap Pay calls, but to skip obvious low-value paid calls when free/local/web sources can answer
  the task well enough. Pay remains the preferred path when a paid provider materially improves the result through freshness, structure, authority, exclusivity, capacity/
  availability, or actionability.

  The guidance also accounts for host-agent capability: agents with reliable built-in browsing/scraping/free API tools can use those for simple public data, while limited
  agents may get a better result from one low-cost structured Pay provider call instead of brittle scraping.

  ## Changes

  - `git diff --check`
  - `cargo fmt --check`
  - `python3 /Users/sarthiborkar/.codex/skills/.system/skill-creator/scripts/quick_validate.py skills/pay`
  - `cargo test -p pay-mcp server_info_has_instructions`

  Note: `just ci` was not run locally because `just` is not installed in this environment.

### DIFF
--- a/rust/crates/core/src/instructions.md
+++ b/rust/crates/core/src/instructions.md
@@ -3,12 +3,33 @@ including HTTP 402, x402, MPP, provider discovery, and wallet-approved calls.
 
 Never answer "Can pay do X" from memory; check `list_catalog`.
 
+# Pay-Worthiness Gate
+
+Before a paid call, do a lightweight check to avoid obvious low-value paid calls.
+
+Use Pay when the task needs live/changing data, structured API responses,
+provider authority, exclusive or rate-limited data, capacity/availability,
+private/provider-owned resources, or an action such as buy, book, send,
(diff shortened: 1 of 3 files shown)


lgalabru:
hey @Sarthib7! 
A ton of energy and manual tests went into these instructions, and as is, the system is a bit fragile.
If you'd like to see the instructions being tweaked, I think the first thing we should do should be implementing some sort of harness to keep track of the triggers we'd like to keep and the ones we want to re-evaluate.
This was explored to an extent in this repo [here](https://github.com/solana-foundation/solana-dev-skill/tree/main/tests), we can something similar in this repo (or find a better approach).

PR #353, opened 2026-05-06
fix: mime type handling

lgalabru:


MavProDev:
You guys are so freaking smart

PR #359, opened 2026-05-08
fix: prevent replay of fulfilled mpp payments

dvansari65:


dvansari65:
- add replay protection for fulfilled MPP payments in the server middleware
- reject duplicate payment proofs with a non-retryable `verification_failed` 402 response
- mark proofs as consumed only after the protected upstream response succeeds
- wire a default replay store into `pay server start` so the protection is active out of the box

lgalabru:
hey @dvansari65! thanks for opening this PR.
This issue was caught and fixed at the `mpp-sdk` level a few days ago.
PR https://github.com/solana-foundation/pay/pull/362 is pulling the latest version of this build which is fixing this issue.
Thanks for looking into this, feel free to re-open if you're seeing something else.

dvansari65:
@lgalabru glad to hear this, while doing this , I thought I am doing wrong cuz I was reading about this 402 payment  security  check on  on some avalanche official page so I thought , I can do something in solana/pay repo.

PR #370, opened 2026-05-25
fix(keystore): build macOS helper for target arch, not build host

### LINKED ISSUE #369: `pay setup` fails on Intel macOS — embedded keychain helper is arm64-only in x86_64 release
## Summary

On Intel Macs, `pay setup` fails when accessing the Keychain because the Swift
helper embedded inside the `x86_64-apple-darwin` release artifact is built as
`arm64`. `posix_spawn`ing it returns `EBADARCH (86)`, surfaced as
`Bad CPU type in executable (os error 86)`.

## Environment

- `@solana/pay` npm: 1.0.16
- `pay` CLI: 0.16.0 (also reproduced on 0.13.0)
- macOS: 15.7.7 Sequoia (Build 24G720)
- CPU: Intel Core i9-8950HK (x86_64)

## Repro

```
npx @solana/pay claude "buy some water with pay"
# or non-interactively:
npx @solana/pay setup --backend keychain --force
```

## Error

```
Error: setup failed: Configuration error: Keystore error: pay.sh: Bad CPU type in executable (os error 86)
```

## Diagnosis

- Main `pay` binary in the x86_64 release is correctly `Mach-O 64-bit executable x86_64`.
- At first run it extracts the embedded Swift helper to `~/.cache/pay/pay.sh`. That file is `Mach-O 64-bit executable arm64`, regardless of host arch.
- Root cause is in `rust/crates/keystore/build.rs`: the `swiftc` invocation passes no `-target` flag, so it builds for the *build host* arch. Release CI appears to run on Apple Silicon, so the helper is always arm64 — even when cross-compiling the main crate for `x86_64-apple-darwin`.

```rust
// rust/crates/keystore/build.rs
let status = std::process::Command::new(SWIFTC)
    .args(["-O", "-o"])
    .arg(&marker)
    .arg(&source)
    .status();
```

## Suggested fix

Either:
1. Pass `-target` to `swiftc` based on `CARGO_CFG_TARGET_ARCH` so the helper matches the crate's target, or
2. Always build a universal helper (`-arch x86_64 -arch arm64`) regardless of host.

Happy to send a PR if a maintainer can confirm preferred approach.

hibari-tech:
## Summary

Fixes #369. The Swift keychain helper was compiled with no `-target`, so `swiftc` defaulted to the build host's architecture. On Apple Silicon CI runners this produced an `arm64` helper even when cross-compiling the crate for `x86_64-apple-darwin` — leaving the released x86_64 `pay` binary with an arm64 helper embedded via `include_bytes!`.

At first run that helper is extracted to `~/.cache/pay/pay.sh` and `posix_spawn`ed, which returns `EBADARCH (86)` ("Bad CPU type in executable") on Intel Macs, breaking `pay setup`.

This passes `-target {CARGO_CFG_TARGET_ARCH}-apple-macos11` to `swiftc` so the helper always matches the crate's target arch (option 1 from #369; happy to switch to a universal helper if preferred).

## Test plan (verified locally on Intel macOS 15.7.7)

- [x] `just lint` (Prettier + ESLint + `cargo clippy --workspace --all-targets -- -D warnings`) — clean
- [x] `cargo fmt --check` — clean
- [x] `cargo build --release --bin pay` — succeeds
- [x] `target/release/build/pay-keystore-*/out/pay-helper` is `Mach-O 64-bit executable x86_64` (was `arm64` before the fix)
- [x] `pay-keystore` unit tests: 47/47 pass
- [x] `rm ~/.cache/pay/pay.sh && pay setup --backend keychain --force` → "Account secured in Apple Keychain" (was `Keystore error: pay.sh: Bad CPU type in executable (os error 86)`)

### Note on `just ci` / `just test`

`cargo test --workspace` is flaky on `main` independently of this PR: `pay-core::server::proxy::tests::forward_request_injects_hmac_auth` and at least one neighbouring test both `set_var`/`remove_var` the same `_TEST_ALIBABA_ACCESS_KEY_SECRET` env var, and `std::env` is process-global, so they race under `cargo test`'s default thread pool and intermittently fail with `HMAC secret env var not set`. The failing test passes deterministically in isolation:

```
cargo test -p pay-core --features server --lib \
  server::proxy::tests::forward_request_injects_hmac_auth
```

The flake is in a different crate from this change (`pay-core` vs `pay-keystore`) and is not introduced by this PR. Happy to file a follow-up if a maintainer confirms the diagnosis.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/rust/crates/keystore/build.rs
+++ b/rust/crates/keystore/build.rs
@@ -34,8 +34,15 @@ fn main() {
             return;
         }
 
+        // Match the helper's CPU arch to the crate's target arch — otherwise
+        // an arm64 build host produces an arm64 helper even when cross-compiling
+        // for x86_64-apple-darwin, which fails with EBADARCH at runtime on Intel.
+        let target_arch =
+            std::env::var("CARGO_CFG_TARGET_ARCH").unwrap_or_else(|_| "x86_64".to_string());
+        let swift_target = format!("{target_arch}-apple-macos11");
+
         let status = std::process::Command::new(SWIFTC)
-            .args(["-O", "-o"])
+            .args(["-O", "-target", &swift_target, "-o"])
             .arg(&marker)
             .arg(&source)
             .status();


lgalabru:
Superseded by https://github.com/solana-foundation/pay/pull/388.
Thanks @hibari-tech!

PR #392, opened 2026-07-01
fix: memo

lgalabru:


EfeDurmaz16 (commented):
## Review — per-challenge memo nonce, sound money path

Verdict: Approve

The client-echoed memo is safely bound to the endpoint resource: `resource_memo_matches` accepts only `resource` (legacy) or `resource#<exactly-3-hex>`, so a client controls just the nonce and cannot stamp another route's memo, change the recipient/amount, or bypass the externalId/memo-mismatch replay check. Legacy static-resource credentials still pass, and the settlement memo is read back from what the client signed rather than re-randomized, so challenge and settlement stay equal.

🟡 Important
- `rust/crates/core/src/server/gate.rs:1265` — `x402_exact_payment_memo` has no unit test while its MPP twin (`mpp_charge_payment_external_id`) has three. This is a reject-legit-payment path: if `accepted.extra.memo` is the wrong location for the signed x402 memo, every x402-exact payment on a resource-backed route is rejected. Add tests mirroring the MPP ones (accepts `resource#hex`, accepts legacy static, rejects wrong resource, rejects non-hex suffix).

💡 Suggestion
- `rust/crates/core/src/server/payment.rs:250` — the nonce is 12 bits (2 random bytes formatted to 4 hex, then `take(3)` discards one hex char), so around 64 same-route payments hit a ~50% memo collision by the birthday bound, which partly defeats the "don't rely on blockhash uniqueness" goal. Keep the full 4 hex (16 bits) or widen the nonce; it costs nothing.

Rust tests green; CLI check and Greptile were still running.

lgalabru:
Addressed the Greptile feedback in a503c86.

- x402 memo extraction no longer pre-rejects credentials when `accepted.extra.memo` is absent or the envelope needs normal SDK validation; it falls back to `process_payment`, while still rejecting an extractable memo that does not match the endpoint resource.
- Added x402 helper coverage for matching memo, wrong-resource memo, missing `accepted.extra.memo`, and missing `accepted` fallback.
- `resource_memo_with_nonce` now keeps the `#abc` nonce inside the protocol memo limit by truncating the resource prefix when needed instead of falling back to a bare non-unique resource; matcher validation now takes the protocol max length.
- PDB only passes `ReceiptLink` into the delivery step when a receipt signature exists, so `SequenceDiagram` does not reserve empty row height for a link that renders `null`.

lgalabru:
Pulled latest and addressed the remaining feedback. CI is green on 93db0e1. The formatter failures were from the suggested guard line; the pushed fix also makes truncated memos hash-bound as prefix#t{hash6}{nonce3}, while keeping normal visible memos as resource#abc. Added the regression guard for api/path#abc vs api/path/extra, and verified cargo fmt --check plus the x402 and MPP memo tests locally. Greptile re-reviewed 93db0e1 and reports safe to merge.

EfeDurmaz16 (commented):
## Re-review — feedback addressed

Verdict: Approve

Both prior points are handled. `x402_exact_payment_memo` now has four tests, and the round-trip helper (`x402_accepted_with_memo` builds the accepted block via `payment_required_header`, then reads it back) confirms `accepted.extra.memo` is the correct location, so the reject-legit-payment risk is closed. The long-resource fallback got a real upgrade: instead of dropping the nonce it now truncates the resource and appends `#t<blake3-6hex><nonce>`, and the matcher rebuilds that deterministically. I traced the new matcher: the memo stays bound to the exact resource via the blake3 hash, `rsplit_once('#')` correctly handles resources containing `#`, truncation is UTF-8-safe, and `max_bytes` is passed consistently on both the challenge and verify sides. The softened extraction (parse error or absent `accepted.memo` now falls back to `Ok(None)` instead of rejecting) is graceful degradation, not a downgrade: the amount/recipient/signature checks still run, so a client omitting the memo gains nothing.

💡 The common-case nonce is still 12 bits (`take(3)` of 4 hex); the improved fallback doesn't change that, so widening it stays a cheap optional win. Not blocking.

EfeDurmaz16 (approved):

PR #393, opened 2026-07-01
fix: unblock release ci and catalog probing

lgalabru:
## Summary
- Tolerate parameterized 404 probe results as warnings so stale resource examples do not block catalog publishing
- Keep PDB dashboard embed tests valid for built Vite output and placeholder output, and make Cargo rerun when PDB dist files change
- Fix GHCR tagging for pay-v* release tags and add docs captured from the paid-path/gaps audit

## Test Plan
- cd rust && cargo fmt --check
- cd rust && cargo clippy --workspace --all-targets -- -D warnings
- cd rust && cargo test --workspace


EfeDurmaz16 (approved):
## Review — tolerate parameterized 404s, unblock release

Verdict: Approve

Release/catalog tooling, no money path. The core change adds a `parameterized_not_found` verdict: a 404 from a probe URL carrying a query string, an unresolved `{path}`, or a `/:param` segment is treated as a warning rather than a hard block, and exempts the provider from the zero-payable-endpoints block, the same "probe can't verify without external state" bucket as credential-gated. The loosening is safe: `is_clean()` now returns false when `tolerated > 0`, so these still surface as CI warnings, and `build_keeps_parameterized_404_as_warning_endpoint` confirms the published endpoint carries empty protocol/USD (unverified, not a claimed payable). The `pay-v*` GHCR tag fix, the recursive `rerun-if-changed` for embedded PDB assets, and the `resolved_endpoint_to_probe_endpoint` dedup are all correct. Good test coverage for the new path.

💡 `has_probe_parameters` tolerates any 404 whose URL has a query string, so a genuinely-removed route probed with `?foo=bar` also lands as a warning rather than a block. Intended per the doc comment, and it still surfaces as a warning, so not blocking.

Verified by line-by-line read (the review workflow stalled today, so I traced it directly).

PR #397, opened 2026-07-08
feat: inference provider as first class citizen

lgalabru:


lgalabru:
@greptile-apps review please.

rust/crates/proxy/src/http402.rs
```diff
@@ -647,21 +753,42 @@ impl<S: PaymentState> ProxyHttp for Http402Gate<S> {
...
-            let _ = self
-                .settle_pending_upto(pending, false, &HeaderMap::new(), None)
-                .await;
+            let served_ok = final_status.is_some_and(|s| (200..300).contains(&s));
+            if let Some(header) = self
+                .settle_pending_upto(pending, served_ok, &HeaderMap::new(), None)
+                .await
+            {
+                deferred_payment_headers.push(header);
+            }
```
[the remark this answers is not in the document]
lgalabru: Addressed with latest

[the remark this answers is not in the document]
lgalabru: @greptile-apps go

lgalabru:
@greptile-apps review please

rust/crates/cli/src/commands/claude/mod.rs
```diff
@@ -63,34 +90,473 @@ impl ClaudeCommand {
...
+        // Direct discovery still supplies the model list for the
+        // ANTHROPIC_DEFAULT_* env vars; the gateway routes by model.
+        Some(choice) if gateway_up => {
+            eprintln!(
+                "⏺ routing claude → gateway {}",
+                inference::LOCAL_GATEWAY_BASE_URL
+            );
+            let upstream = payer_upstream(
+                &choice.provider,
+                inference::LOCAL_GATEWAY_BASE_URL.to_string(),
```
EfeDurmaz16: When the gateway is already up, routing through plain `127.0.0.1:1402` drops the selected provider host, so multi-provider gateways cannot resolve `ollama`/`vllm`/etc. This should target the selected `{slug}.localhost` gateway URL or otherwise preserve the provider `Host`.

rust/crates/cli/src/commands/server/inference/spec.rs
```diff
@@ -0,0 +1,401 @@
...
+        .map(|(model, rate)| MeterVariant {
+            param: "model".to_string(),
+            value: model.clone(),
+            description: None,
+            dimensions: token_dimensions(rate),
+        })
+        .collect();
+
+    // Fallback dimensions: the config `default` rate when no variant matches.
+    let dimensions = config
```
EfeDurmaz16, line 113: These default dimensions are currently bypassed when variants exist: core metering falls back to the first variant if no model variant matches. A model covered only by `default` can therefore settle at the first explicit model's rate instead of the configured default.

EfeDurmaz16 (changes requested):
I reviewed the inference-provider changes end to end and the overall shape is promising: provider discovery, catalog-backed providers, PDB visibility, and token-metered specs fit the existing architecture. I found two behavior issues that are worth fixing before merge. The Claude path can drop the selected provider host when routing through an already-running gateway, and the synthesized default token pricing is currently bypassed by core variant resolution. I’m requesting changes for those two items; the targeted Rust inference/proxy tests I ran passed.

rust/crates/cli/src/commands/server/inference/spec.rs
lgalabru: Addressed with latest

rust/crates/cli/src/commands/claude/mod.rs
lgalabru: Addressed with latest

rust/crates/cli/src/commands/server/inference/mod.rs
```diff
@@ -0,0 +1,1108 @@
...
+        // Per-model pricing: validate against the models each provider
+        // actually serves, then overlay it so the picker can show per-model
+        // in/out token rates. Fail BEFORE opening the gateway/TUI, the same
+        // way the sandbox guard does — a typo'd model or a provider that
+        // serves none of the configured models (with no `default`) is a loud
+        // error, not a silently-dropped chip.
+        if let Some(config) = &pricing_config {
+            let mut errs: Vec<String> = Vec::new();
+            for provider in &discovered {
+                errs.extend(config.validate(provider.slug(), &provider.models));
```
EfeDurmaz16: This validates every model-specific price against each discovered provider, so a valid Ollama-only model entry can fail startup just because LM Studio/vLLM is also discovered. Please validate per-model keys against the union of discovered models, or only require each provider to have at least one resolved rate/default.

EfeDurmaz16 (changes requested):
Thanks for the updates here. I re-reviewed current head eda5df1 and the earlier payer-proxy receipt concern looks fixed: translated JSON, SSE, MPP retry, and x402-upto retry paths preserve PAYMENT-RESPONSE, and the focused payer-proxy tests pass locally. The inference/metering path also has good coverage, but I found one startup blocker in model-specific pricing validation for multi-provider setups. Verdict: request changes for that validation issue; CI otherwise looks green from the PR rollup.

EfeDurmaz16 (approved):
Re-reviewed current head 6e8c3f4. The provider-host routing, translated payment receipt paths, and pricing validation issues I previously flagged are fixed; the new validation checks model keys against the discovered model union while still requiring each routed provider to resolve a rate/default. Focused Rust tests pass locally for pricing validation, provider-host routing, translated JSON/SSE/MPP/x402-upto receipt handling, and priced specs; the PR checks are green. Verdict: approve.

PR #398, opened 2026-07-11
fix: linux headless, pay-kit update

lgalabru:


rust/crates/core/src/client/runner.rs
```diff
@@ -402,15 +439,16 @@ pub(crate) fn parse_httpie_output(
...
             continue;
         }
 
         if in_response_headers {
             if trimmed_full.is_empty() {
-                body_lines = Some(lines[(i + 1)..].to_vec());
-                break;
+                body_start = Some(i + 1);
+                in_response_headers = false;
+                continue;
```
EfeDurmaz16: 🔴 blocking: Once the real response body starts, a line beginning `HTTP/` must remain body. Continuing here lets a 200 body be re-parsed as a 402 with attacker-controlled payment headers and routes HTTP-tool requests through the auto-pay authorization path. Use unambiguous history framing and add a 200-with-fake-402 regression.

EfeDurmaz16 (commented):
## Review: HTTPie response parsing can manufacture a payment challenge

Verdict: Not approve.

CI is green, but the HTTPie parser can reinterpret attacker-controlled content from a real 200 response as a synthetic 402 and pass it into the normal payment authorization path.

Two non-blocking follow-ups:
- Make file-account installation atomically no-clobber so concurrent creates cannot overwrite the first private key.
- Surface Secret Service relock failures after secret loads instead of returning the secret while leaving a previously locked collection open.

I reproduced the blocker with a focused regression test. The full `pay-core` and `pay-keystore` test suites pass otherwise.

rust/crates/cli/src/commands/mod.rs
```diff
@@ -251,17 +263,34 @@ impl Command {
...
             Command::Fetch(cmd) => {
-                let parsed_headers = parse_header_args(&cmd.headers);
-                pay_core::skills::validate_cached_catalog_request("GET", &cmd.url, None)?;
-                let outcome = pay_core::fetch::fetch(&cmd.url, &parsed_headers)?;
-                let tool = Tool::Fetch { url: &cmd.url };
+                let prepared = cmd.prepare()?;
+                pay_core::skills::validate_cached_catalog_request(
+                    &prepared.method,
+                    &cmd.url,
+                    prepared.validation_body.as_deref(),
```
EfeDurmaz16: 🔴 blocking: `None` means no body to this validator, but multipart and binary uploads reach here with a populated `RequestBody` and `validation_body=None`. Matching cached endpoints with a required request body are rejected before sending, so please preserve a body-present/non-JSON state or validate method and query separately, with a cached multipart regression test.

rust/crates/mcp/src/tools/curl.rs
```diff
@@ -119,6 +184,156 @@ pub async fn run(
...
+        })
+    });
+    if !permitted {
+        return Err(format!(
+            "`body_file` `{}` is outside the MCP client's declared filesystem roots.",
+            canonical_file.display()
+        ));
+    }
+
+    let (body, content_type) = RequestBody::from_file(&path)
```
EfeDurmaz16: 🔴 blocking: Root containment is authorized on `canonical_file`, but this reopens the original path. A writable ancestor can be swapped after `canonicalize`, causing Pay to snapshot an outside-root file; `O_NOFOLLOW` protects only the final component on Unix, and Windows also follows a swapped final reparse point. Please open once, validate the opened target against MCP roots, and read that same handle.

rust/crates/cli/src/components/ascii_table.rs
```diff
@@ -0,0 +1,116 @@
...
+        output.push(' ');
+        output.push_str(&cell);
+        output.push_str(&" ".repeat(widths[index].saturating_sub(display_width(&cell)) + 1));
+        output.push('|');
+    }
+    output.push('\n');
+}
+
+fn abbreviate(value: &str, width: usize) -> String {
+    let value = value.replace(['\r', '\n'], " ");
```
EfeDurmaz16: 🔴 blocking: These cells contain decoded server JSON, so replacing only CR/LF leaves ESC, BEL, and other terminal controls executable under `--verbose`. Strip or escape all C0/C1 controls before width calculation and rendering, and apply the same rule to receipt titles and OSC-8 link components.

rust/crates/cli/src/commands/mod.rs
```diff
@@ -742,8 +779,12 @@ fn handle_outcome(
         }
 
         RunOutcome::Completed {
-            exit_code, body, ..
+            exit_code,
+            body,
+            response_headers,
+            ..
         } => {
+            print_verbose_receipt(&response_headers, network_override, None, verbose, is_json);
```
EfeDurmaz16: 🔴 blocking: An initial unpaid 200 also reaches this call, so server-supplied receipt headers are presented as `Payment completed` or `... paid` even though Pay never authorized or retried. Gate paid wording and receipt rendering on explicit paid-retry provenance, with an initial-200 negative test.

EfeDurmaz16 (commented):
## Review: not ready to approve

The new one-shot snapshots address the same-size replacement race, and the latest commit fixes inline body content types. CI is green, but the remaining payment, upload, and terminal-output blockers are detailed inline.

Focused verification reproduced the HTTPie 200-to-402 misclassification and terminal-control output on the current head. The normal suites still pass: pay 401/401, pay-core fetch 19/19, and pay-mcp 104/104. The three MCP elicitation E2Es are all ignored, which is why these upload boundary failures remain false-green.

Also important: `body_file` is fully snapshotted before elicitation even though the tool description and prompt say it is read after approval. Please align the implementation with the consent contract and add accept, decline, cancel, and timeout coverage. Ping me after the fixes and I will re-review.

rust/crates/core/src/client/runner.rs
EfeDurmaz16: Re-review on d658758: this still reproduces. A focused external-client test returns `X402Challenge` for a genuine 200 body containing an embedded 402 transcript, and this parser is unchanged since b893382; please keep this thread open until the regression is added.

lgalabru: Fixed in b2925ee. HTTPie now forces --no-all and the parser stops at the first response header terminator, so embedded HTTP-looking response-body text cannot create a synthetic 402. Added a regression covering a 200 body containing a fake 402 transcript.

rust/crates/cli/src/components/ascii_table.rs
lgalabru: Fixed in 9aa6cdb. Added shared C0/DEL/C1 sanitization before table sizing/rendering and receipt-title display; OSC-8 link labels and targets are sanitized before escape sequences are generated. Added regression coverage for table cells, receipt titles, and links.

rust/crates/cli/src/commands/mod.rs
lgalabru: Fixed in ed4d596. Completion paths now carry explicit provenance: initial responses never render receipt/payment messaging, while only a completed paid retry may render a receipt. Added an initial-200-with-receipt-header negative test.

rust/crates/cli/src/commands/mod.rs
lgalabru: Fixed in a6e01f9. Opaque multipart/binary bodies now retain cached method/path/query validation but intentionally skip JSON-schema body validation, instead of being treated as absent. Added a regression proving required JSON bodies do not block opaque uploads while required query validation still applies.

lgalabru:
Addressed the elicitation consent-contract concern in 2025636: body_file bytes are now read only after the user accepts. Pay snapshots immediately after approval and rejects the request if the approved file changed before sending; the focused changed-file regression passes.

rust/crates/mcp/src/tools/curl.rs
lgalabru: Fixed in d0eac20. Pay now records the approved file identity before elicitation; after approval it opens once, verifies that exact opened handle still identifies the approved file, and snapshots bytes from that same handle. The Unix regression swaps the approved root for an outside-root symlink and confirms the upload is rejected.

rust/crates/core/src/client/runner.rs
lgalabru: Resolved by b2925ee; the parser regression now covers a genuine 200 body containing an embedded 402 transcript and preserves it as body text.

rust/crates/cli/src/commands/mod.rs
```diff
@@ -1460,11 +1905,27 @@ fn retry_header_args_httpie(headers_to_add: &[(&str, String)]) -> Vec<String> {
...
-            exit_code, body, ..
+            exit_code,
+            body,
+            response_headers,
+            ..
         } => {
+            print_verbose_receipt(
+                &response_headers,
+                receipt_network,
+                ReceiptProvenance::PaidRetry(payment),
```
EfeDurmaz16: Blocking: This helper is also used by the x402 SIWX sign-in path at line 1625, so a non-payment authentication retry is treated as a paid retry here. If the sign-in response includes `payment-receipt-url`, verbose output can still render `Payment completed`; please pass explicit provenance from each caller and add a sign-in negative regression.

rust/crates/core/src/skills/openapi.rs
```diff
@@ -1214,7 +1235,12 @@ fn validate_openapi3_request(
     let spec_path = normalize_path(spec_path);
     let mut query_problems =
         validate_openapi_query_parameters(doc, item_obj, operation, query_params);
-    match validate_openapi_operation_body(doc, method, &spec_path, operation, body) {
+    let body_outcome = if validate_body {
```
EfeDurmaz16, line 1256: Blocking: Skipping JSON schema validation should not skip the whole `requestBody` contract. With `validate_body=false`, opaque bodies are accepted even when the operation declares no body or only a different media type, and non-UTF8 bytes labelled `application/json` can reach the same path; please validate body presence and effective Content-Type before signing, then only skip schema decoding.

rust/crates/mcp/src/tools/curl.rs
```diff
@@ -119,6 +187,225 @@ pub async fn run(
...
+        ));
+    }
+
+    let modified = metadata.modified().map_err(|error| {
+        format!(
+            "Could not read modification time for `body_file` `{}`: {error}",
+            path.display()
+        )
+    })?;
+    let handle = same_file::Handle::from_path(&canonical_file).map_err(|error| {
```
EfeDurmaz16, line 320: Blocking: Containment is checked on `canonical_file` before this separate path lookup. A writable ancestor can be swapped to an outside-root symlink or junction before `Handle::from_path` opens it, so the expected handle can already identify the outside file and the later identity checks pass. Please bind root containment and reading to one no-follow handle and add a regression that swaps before the first handle opens.

EfeDurmaz16 (commented):
Review on 048a114: not approve yet.

The HTTPie boundary parsing, terminal sanitization, initial-response receipt gating, approval-before-read, and post-approval snapshot reuse fixes are materially better, and the focused regressions I ran locally pass.

I still see three blocking variants:
- x402 SIWX sign-in retries still reuse paid-retry receipt provenance.
- Opaque fetch body validation now skips the full requestBody contract, not only JSON schema decoding.
- MCP `body_file` root containment still has a pre-handle race before the retained identity is opened.

CI is green on GitHub, but these are payment/upload authorization boundaries, so I would keep this in review until they are fixed.

rust/crates/cli/src/commands/mod.rs
lgalabru: Fixed in dbe7cce. Completion provenance now distinguishes non-payment retries from paid retries; SIWX uses NonPaymentRetry, so receipt headers cannot render payment messaging. Added a SIWX-style negative regression.

rust/crates/cli/src/commands/mod.rs
```diff
@@ -1354,14 +1763,36 @@ fn pay_session_and_retry(
...
         Tool::Curl(args) => pay_core::runner::validate_curl_args_against_catalog(args),
-        Tool::Fetch { url } => pay_core::skills::validate_cached_catalog_request("GET", url, None),
+        Tool::Fetch {
+            method,
+            url,
+            validation_body,
+            ..
+        } if body.is_some() && validation_body.is_none() => {
+            pay_core::skills::validate_cached_catalog_request_metadata(method, url)
+        }
```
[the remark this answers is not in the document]
lgalabru: Already fixed in 048a114. The guarded Tool::Fetch pattern now explicitly binds body, and CI passed the workspace clippy compile after that commit.

rust/crates/core/src/skills/openapi.rs
lgalabru: Fixed in 979cafa. Opaque bodies now retain method/query validation and must match an operation-declared requestBody media type; only JSON schema decoding is skipped. Operations with no requestBody or a different Content-Type are rejected before signing.

rust/crates/core/src/skills/openapi.rs
```diff
@@ -1138,14 +1138,51 @@ pub fn validate_request(
...
+/// Validate a request's method, path, and query parameters while optionally
+/// skipping JSON-schema validation for an opaque, non-JSON request body.
+pub fn validate_request_with_body_validation(
+    doc: &Value,
+    method: &str,
+    relative_path: &str,
+    query_params: &[(String, String)],
+    body: Option<&str>,
+    validate_body: bool,
+    opaque_content_type: Option<&str>,
```
EfeDurmaz16, line 1162: Blocking: This signature now requires `opaque_content_type`, but the existing test callers at lines 3122 and 3133 were not updated. The crate no longer compiles under `cargo test` with E0061, so please update the callers and keep the opaque media regressions running in CI.

EfeDurmaz16 (commented):
## Review - 979cafa re-review

Verdict: Not approve

This head does not compile. The SIWX provenance fix looks directionally correct, but the OpenAPI helper signature change left existing test callers on the old arity.

Blocking
- `rust/crates/core/src/skills/openapi.rs:1162` - `validate_request_with_body_validation` now takes `opaque_content_type`, but callers at `openapi.rs:3122` and `openapi.rs:3133` still pass 6 arguments. `cargo test -p pay-core opaque_body_validation_keeps_method_and_query_validation -- --nocapture` fails with E0061, and GitHub CI now shows Rust lint, Rust tests + coverage, and CLI check failing.

Still open
- The previous MCP `body_file` root-containment blocker is unchanged in this delta, so I did not add a duplicate inline.

rust/crates/mcp/src/tools/curl.rs
```diff
@@ -830,6 +1105,50 @@ fn do_paid_fetch(
...
+        return pay_core::skills::validate_cached_catalog_request(method, url, None);
+    };
+
+    let content_type = headers
+        .iter()
+        .find(|(name, _)| name.eq_ignore_ascii_case("content-type"))
+        .map(|(_, value)| value.as_str());
+    if !content_type.is_some_and(is_json_media_type) {
+        return Ok(());
+    }
```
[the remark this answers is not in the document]
lgalabru: Fixed in 668561e. Non-JSON MCP uploads now use `validate_cached_catalog_opaque_request`, preserving cached method, path, query, and declared media-type validation while skipping only JSON-schema parsing.

EfeDurmaz16 (commented):
## Re-review - 668561e

Verdict: Not ready to approve

The previous OpenAPI compile blocker is fixed, and opaque MCP uploads now retain cached method, query, and exact media-type validation. All CI checks are green, and the focused OpenAPI and MCP regression tests pass locally.

### Blocking

- The previously reported `body_file` root-containment race remains unchanged at `rust/crates/mcp/src/tools/curl.rs:320`. Containment is checked before the first path open. A writable ancestor can therefore be swapped before `Handle::from_path`, causing the retained expected handle and the post-approval reopen to both resolve to the same file outside the declared MCP root. The existing `swapped_root_cannot_redirect_an_approved_file` test swaps only after the expected handle has already been captured, so it does not cover this interleaving. Existing thread: https://github.com/solana-foundation/pay/pull/398#discussion_r3589247595

I am leaving this as a comment rather than an approval until containment and reading are bound to a handle acquired without this second path lookup, with a regression for the pre-handle swap window.

rust/crates/mcp/src/tools/curl.rs
EfeDurmaz16: Reassessed against the base branch and the practical prerequisites: I am downgrading this from blocking to a non-blocking hardening follow-up. The race is real, but the base branch currently permits unrestricted `body_file` reads, while this PR adds MCP-root checks, user approval, and file-identity verification. Exploitation of the remaining window requires a concurrent local actor, a narrow timing win, and matching the approved file's exact size and modification time. Worth addressing separately, but it should not hold this security improvement.

EfeDurmaz16 (approved):
## Re-review - final

Verdict: Approve

I re-calibrated the previously reported pre-handle race against the base branch and its practical attack prerequisites. It is a valid hardening follow-up, but not proportionate to block this PR: exploitation requires a concurrent local actor to win a narrow window and preserve the approved file's exact size and modification time, while the base branch currently performs unrestricted `body_file` reads without MCP-root checks or user elicitation. This PR materially improves that boundary.

The compile regression is fixed, opaque uploads retain cached request validation, all CI checks are green, and the focused OpenAPI and MCP tests pass.

Issue #395, opened 2026-07-07
gateway-402: generateContent/streamGenerateContent return 500 challenge_generation_failed on the Gemini host


eladmint:
### Context

We're Sippar (production consumer, same account behind #394 and pay-kit#212/#213). We pay `generativelanguage.google.gateway-402.com` server-side with `@solana/pay-kit`'s `client.fetch`, settling real USDC on mainnet — this has worked since 2026-07-04 (our `generateContent` vision-critic call, proven live).

As of today (2026-07-07) the same call fails before any payment step.

### Problem

`generateContent` and `streamGenerateContent` on this host return a bare `500` with no 402 at all:

```
$ curl -s -i -X POST "https://generativelanguage.google.gateway-402.com/v1beta/models/gemini-2.5-flash:generateContent" \
  -H "Content-Type: application/json" \
  -d '{"contents":[{"role":"user","parts":[{"text":"hi"}]}]}'

HTTP/2 500
content-type: application/json

{"error":"challenge_generation_failed"}
```

Reproduced consistently (30+ calls) across:
- two networks (our own machine and our production VPS)
- curl and Node's native `fetch`
- every model we tried (`gemini-2.5-flash`, `gemini-2.5-pro`, `gemini-2.0-flash`, `gemini-1.5-flash`)
- both `generateContent` and `streamGenerateContent`
- with and without a `tools`/`toolConfig` body, with and without a body-level `model` field, with vision (`inlineData`) and text-only content

Meanwhile the same host's `embedContent` and `countTokens` endpoints work correctly right now — `embedContent` returns a clean 402 with a full MPP challenge, `countTokens` returns 200 with no payment required. So the gateway and its payment stack are up; this looks specific to the two content-generation methods.

### Why we think this is new, not a known limit

- #394 (which we filed 2026-07-02) shows `generateContent`'s `x402-upto` challenge decoding fine that day.
- Our own `generateContent` vision-critic call has been live in production since 2026-07-04.

So this is a regression inside the last few days, not a documented constraint.

### Possible cause (unconfirmed, just a lead)

The proxy was rewritten onto Pingora in #385 (merged 2026-06-29). That PR's own "Deferred" section lists "response-side session byte-metering" as not yet done. `generateContent`/`streamGenerateContent` are the two methods whose price depends on the *output* size, which is unknown until generation completes — unlike `embedContent`'s flat, input-only pricing. That split lines up with what we're seeing, but we haven't looked at the proxy internals to confirm it.

[...]

PR #399, opened 2026-07-13
Add x402-upto challenge regression test for Gemini generateContent

koriyoshi2041:
## Problem

`#395` covered a gateway path where Gemini `generateContent`/`streamGenerateContent` could fail before returning a payable challenge. The live endpoint now returns `402`, but there was no focused core gate regression for the `x402-upto` challenge boundary.

## Fix

Add a test-only Gemini-shaped `generateContent` route that accepts only `x402-upto`, then assert an unauthenticated request returns `402 payment_required`, advertises `payment.schemes == ["x402-upto"]`, and includes the challenge header. The test uses an offline cached blockhash, so it does not require live RPC, wallet funding, or settlement.

## Test

- `cargo fmt -p pay-core`
- `cargo test -p pay-core --features server gemini_generate_content_x402_upto_challenges_without_payment`
- `git diff --check`

## Risk

Low. This is test-only coverage in the core gate path and should not change runtime payment behavior.

### DIFF
--- a/rust/crates/core/src/server/gate.rs
+++ b/rust/crates/core/src/server/gate.rs
@@ -1426,6 +1426,8 @@ async fn session_authorized(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use pay_kit::mpp::solana_keychain::{SolanaSigner, memory::MemorySigner};
+    use std::sync::Arc;
 
     // Ceiling $0.10 at 6 decimals == 100_000 base units (USDC).
     const CEILING_USD: f64 = 0.10;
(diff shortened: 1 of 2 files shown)


lgalabru (dismissed):
Thanks @koriyoshi2041!

koriyoshi2041:
Revalidated the current head after the clippy cleanup: `cargo fmt --all -- --check`, `cargo clippy -p pay-core --all-targets -- -D warnings`, and the focused `pay-core` Gemini challenge test all pass. The branch is conflict-free; the only reported check is the fork Vercel deployment authorization. Ready for another look when convenient.

PR #402, opened 2026-07-16
fix: windows build for pay-proxy shutdown

lgalabru:
Pingora 0.5's caller-driven shutdown (RunArgs::shutdown_signal) and tokio::signal::unix are unix-only, so run_with_shutdown/WatchShutdown failed to compile on *-pc-windows-msvc. cfg-gate the unix-only pieces and fall back to RunArgs::default() on Windows.

### DIFF
--- a/rust/crates/proxy/src/lib.rs
+++ b/rust/crates/proxy/src/lib.rs
@@ -13,10 +13,14 @@ pub mod observer;
 
 pub use http402::Http402Gate;
 
-use async_trait::async_trait;
 use pay_core::PaymentState;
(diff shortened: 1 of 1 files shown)


EfeDurmaz16 (approved):

EfeDurmaz16:
LGTM, one quick question:

Greptile review shows 

The Unix host cargo check was attempted but failed before the libz-ng-sys v1.1.29 validation, reporting No such file or directory (os error 2) and hints that cmake may not be installed, exit code 101.

 @lgalabru should we look into this?

PR #405, opened 2026-07-21
build: update PayKit dependency

lgalabru:
## Summary
- Update the PayKit git dependency to merged PayKit main (76402a5f).
- Pull in delegated session signer enforcement, RPC retry hardening, and the public settlement-authority wire type.

## Test Plan
- cargo check -p pay --all-features
- Greptile review and GitHub CI


rust/crates/proxy/src/http402.rs
```diff
@@ -450,6 +553,76 @@ impl<S: PaymentState> Http402Gate<S> {
...
+                // has already been produced, so preserve it and leave this
+                // over-cap response unsettled rather than turning it into a
+                // misleading gateway failure.
+                pending.handle.touch_channel(pending.channel_id.clone());
+                tracing::warn!(
+                    channel = %pending.channel_id,
+                    error = %error,
+                    "delegated MPP response exceeded remaining session capacity"
+                );
+                return Ok(None);
```
EfeDurmaz16: Every `authorize_delegated_usage` error is currently converted into a free success response. Reserve channel capacity atomically before forwarding (`committed + reserved <= deposit`), then commit actual usage or release the reservation; a signer-side mutex at settlement is too late.

rust/crates/proxy/src/http402.rs
```diff
@@ -568,20 +746,35 @@ impl<S: PaymentState> ProxyHttp for Http402Gate<S> {
...
+                    .and_then(|pending| pending.settlement.as_ref())
+                    .is_some_and(|plan| {
                         metering::upto_requires_response_body(
                             &plan.metering,
                             plan.variant_hint.as_deref(),
                         )
-                    })
-                }) {
+                    });
+                if delegated_session || response_metered_upto {
```
EfeDurmaz16, line 669: `UpstreamPlan::Respond` reaches `finish_buffered_axum_response` without delegated settlement. Metered delegated `routing: respond` endpoints are served for free.

rust/crates/proxy/src/http402.rs
```diff
@@ -407,7 +483,29 @@ impl<S: PaymentState> Http402Gate<S> {
             }
         };
 
-        self.observe_buffered_upto_response(ctx, &response_headers, &body);
+        self.observe_buffered_metered_response(ctx, &response_headers, &body);
+        if status.is_success() {
+            match self
+                .settle_delegated_session(ctx, &response_headers, &body)
```
EfeDurmaz16, line 429: Treat the durable delivery reservation as the in-flight lease and block auto-close while it is unresolved; release it only on commit or terminal failure. The deadline-only touch can close a channel under a slow request.

rust/crates/core/src/server/gate.rs
```diff
@@ -1347,24 +1363,93 @@ fn variant_hint_from_path(path: &str) -> Option<String> {
...
-            session: handle.map(|h| SessionForward {
-                handle: h,
-                channel_id: state.channel_id,
-                committed_base_units: state.cumulative,
-            }),
-            receipt: None,
-            upto: None,
-        },
+        Ok(SessionOutcome::Active { state, signature }) => {
+            if sm.settlement_authority() == pay_kit::mpp::SessionSettlementAuthority::ClientVoucher
```
EfeDurmaz16, line 1505: Keeping `SessionHandle` would not be a safe fix: this 402 supplies no resource-bound amount, and `minVoucherDelta` is only a floor. Return and bind a required cumulative amount before retrying; otherwise auto-pay exits while a custom client can arbitrarily underpay.

rust/crates/core/src/server/gate.rs
```diff
@@ -1347,24 +1363,93 @@ fn variant_hint_from_path(path: &str) -> Option<String> {
...
+                request_properties: props,
+                ceiling_usd,
+                inferred_usage: None,
+            }));
+            GateDecision::Forward {
+                session: Some(SessionForward {
+                    handle,
+                    channel_id: state.channel_id,
+                    committed_base_units: state.cumulative,
+                    settlement,
```
EfeDurmaz16, line 1567: The delegated settlement plan is ignored by `payment_middleware`; only Pingora consumes it. Axum users can return delegated responses without a debit.

rust/crates/cli/src/commands/server/start.rs
```diff
@@ -955,49 +1016,66 @@ impl StartCommand {
                 } else {
                     None
                 };
-                let session_operator = session_channel_payer_signer
+                let session_operator = fee_payer_signer
```
EfeDurmaz16, line 1019: Sandbox client-voucher pulls advertise `fee_payer_signer` but validate against `session_channel_payer_signer`. When they differ, the client-built open is rejected.

rust/crates/cli/src/commands/server/provider_registration.rs
```diff
@@ -0,0 +1,470 @@
...
+        }
+        eprintln!(
+            "⏺ renewing {} / {} as {}\n  endpoint: {}\n  authority: {}\n  registry PDA: {}\n  RPC: {}",
+            registration.category,
+            registration.protocol,
+            registration.name,
+            registration.endpoint,
+            authority,
+            pda,
+            rpc_url,
```
EfeDurmaz16, line 142: Raw RPC URLs are written to stderr. Redact userinfo, path, and query before logging.

rust/crates/cli/src/commands/server/inference/mod.rs
```diff
@@ -399,6 +408,50 @@ impl InferenceCommand {
...
+                );
+                Arc::new(pay_core::signer::load_signer_with_intent(source, &intent)?)
+                    as Arc<dyn SolanaSigner>
+            };
+            let registration = super::provider_registration::ServiceRegistration::new(
+                "inference",
+                "openai-v1",
+                "pay-inference",
+                "OpenAI-compatible inference served through pay",
+                endpoint,
```
EfeDurmaz16, line 433: This publishes the bare gateway URL, but multi-provider routing only accepts `{slug}.<gateway>` and the bare host returns 404. The current PDA identity permits only one entry per authority/category/protocol, so add a default route or extend the registry identity before publishing provider-specific URLs.

rust/crates/core/src/client/runner.rs
```diff
@@ -685,15 +685,16 @@ pub(crate) fn classify_402_with_preference(
     // the recipient is a valid Solana pubkey.
     // Session MPP: the method field ("solana") indicates chain support.
     // Session requests don't use ChargeRequest so mpp_is_solana doesn't apply.
-    if let Some(challenge) = mpp_challenges
-        .iter()
-        .find(|challenge| challenge.intent.as_str() == "session")
+    if preference != ProtocolPreference::OnlyX402
```
EfeDurmaz16, line 688: The server now advertises multiple session currencies, but the client selects only the first challenge. Later funded currencies are unreachable.

rust/crates/cli/src/commands/server/provider_registration.rs
```diff
@@ -0,0 +1,470 @@
...
+//! discovery. PDA seed bytes themselves are hashed and cannot be filtered.
+
+use std::str::FromStr;
+use std::sync::Arc;
+
+use pay_kit::mpp::solana_keychain::SolanaSigner;
+use solana_instruction::{AccountMeta, Instruction};
+use solana_pubkey::Pubkey;
+
+pub const DEFAULT_PROGRAM_ID: &str = "FJwLLb2G63XfXs43vQRYtXYXXM515d9UGt3paNLtm6j5";
```
EfeDurmaz16, line 14: This program ID is currently absent on mainnet, devnet, and Surfnet. Since registration blocks startup, deploy it or make registration nonfatal before enabling it by default.

EfeDurmaz16 (changes requested):
Delegated session settlement can fail open across Pingora and Axum, and the default client-voucher path cannot complete. The remaining comments cover concrete routing, signer, secret-handling, and deployment blockers. Requesting changes.

rust/crates/core/src/server/gate.rs
```diff
@@ -1347,24 +1363,93 @@ fn variant_hint_from_path(path: &str) -> Option<String> {
...
+                    available_base_units,
+                }),
+                receipt: signature
+                    .map(|reference| session_receipt_annotation(sm.network(), reference)),
+                upto: None,
+            }
+        }
         Ok(SessionOutcome::Voucher {
```
EfeDurmaz16, line 1576: Voucher verification checks signature, monotonicity, cap, and the minimum delta, then this branch forwards with `settlement: None`. The added test accepts a 25-unit voucher for a 1,000-unit endpoint; reject vouchers below a resource-bound quote.

lgalabru:
@EfeDurmaz16 I addressed the reservation concern in commits 64fcc50 and b082f92: delegated capacity is reserved atomically before forwarding and released on settlement or terminal response/failure. Delegated settlement is buffered for Axum and Pingora before successful responses are released.\n\nI agree with the remaining correctness concerns:\n- client-voucher requests need a resource-bound quote; minVoucherDelta alone is insufficient\n- sandbox signer validation needs to use the effective settlement signer\n- RPC URLs need redacted logging\n- provider registration must be gated/nonfatal until routing and program deployment are available\n\nFor multi-currency client selection, I’m pushing back on silently choosing a later currency: this needs an explicit wallet/client preference API.

rust/crates/core/src/server/session.rs
```diff
@@ -81,6 +88,21 @@ struct SessionOperatorRuntime {
...
 
 impl SessionOperatorRuntime {
+    fn reserve_capacity(&self, channel_id: &str, amount: u64) -> bool {
+        let Ok(mut reservations) = self.reserved_capacity.lock() else {
+            return false;
+        };
+        if reservations.contains_key(channel_id) {
+            return false;
+        }
+        reservations.insert(channel_id.to_string(), amount);
```
EfeDurmaz16, line 98: Could this reservation also defer lifecycle close? `close_due_channels` never checks it, so a slow in-flight response can still seal the channel before settlement.

rust/crates/proxy/src/http402.rs
```diff
@@ -450,6 +558,74 @@ impl<S: PaymentState> Http402Gate<S> {
...
+        let Some(plan) = pending.settlement else {
+            return Ok(None);
+        };
+        let actual = metering::upto_actual_amount_from_response(
+            &plan,
+            pending.available_base_units,
+            response_headers,
+            Some(response_body),
+        )
+        .map_err(|error| error.to_string())?;
```
EfeDurmaz16: `upto_actual_amount_from_response(...)?` can return after `ctx.session.take()` but before any release, leaving the channel permanently reserved. A scoped/idempotent lease guard would cover this and the earlier terminal exits.

EfeDurmaz16 (commented):
Thanks for addressing atomic admission and the fail-closed settlement path; those changes fix the concurrent free-response case. I still see two reservation lifecycle gaps below, and the existing Axum/respond and client-voucher findings remain applicable.

Sorry for the back and forth. Happy to do these as a follow-up PR if you'd like.

EfeDurmaz16 (approved):
LGTM, thanks for the changes! The delegated settlement path now fails closed and the reservation lifecycle looks solid. The remaining client-voucher amount binding can be handled in a focused follow-up PR.

PR #406, opened 2026-07-22
feat(cli): add paid agent provider routing

lgalabru:
## Summary
- update PayKit with durable Redis-backed channel state and TLS support
- add paid Alibaba Model Studio and Gemini routing for Claude, Codex, Goose, and Qoder
- recover cached hosted sessions by rediscovering and opening a replacement channel
- make idle close omit vouchers whose watermark already landed, avoiding 0xEA duplicate settlement failures
- leave periodic durable watermark reconciliation to the singleton Agent Gateway job

## Test Plan
- cargo fmt --all -- --check
- payer proxy replacement-session regression test
- close watermark comparison regression test
- PayKit Redis store tests, including concurrent watermark CAS
- Agent Gateway settle-sessions check, clippy, test build, and Terraform validation

### DIFF
--- a/rust/crates/cli/src/commands/goose.rs
+++ b/rust/crates/cli/src/commands/goose.rs
@@ -1,0 +1,193 @@
+//! `pay goose` — launch Goose with the pay MCP server and optional paid
+//! alternate-provider routing.
+
+use std::process::{Command, Stdio};
+
+use clap::Args;
+
+use super::claude::{AlternateClient, AlternateProvider, prepare_alternate_provider};
+
+const GOOSE_PROVIDER_ENV: &str = "GOOSE_PROVIDER";
+const GOOSE_MODEL_ENV: &str = "GOOSE_MODEL";
+const OPENAI_HOST_ENV: &str = "OPENAI_HOST";
+const OPENAI_API_KEY_ENV: &str = "OPENAI_API_KEY";
+const OPENAI_BASE_PATH_ENV: &str = "OPENAI_BASE_PATH";
+const GOOSE_DISABLE_SESSION_NAMING_ENV: &str = "GOOSE_DISABLE_SESSION_NAMING";
+
+/// Run Goose with 402 payment support.
+#[derive(Args)]
+#[command(disable_help_flag = true)]
+pub struct GooseCommand {
+    /// Arguments forwarded to `goose session`.
+    #[arg(trailing_var_arg = true, allow_hyphen_values = true)]
+    pub args: Vec<String>,
+}
+
+impl GooseCommand {
+    pub fn run(
+        self,
+        pay_bin: &str,
+        active_account_name: Option<&str>,
+        network_override: Option<&str>,
+        _alternate_provider: bool,
+    ) -> pay_core::Result<i32> {
+        if goose_version_requested(&self.args) {
+            return launch_goose(&self.args, None, active_account_name);
+        }
+        if goose_help_requested(&self.args) {
+            let mut args = vec!["session".to_string()];
+            args.extend(self.args);
+            return launch_goose(&args, None, active_account_name);
+        }
+
+        // Goose is intentionally alternate-first: both `pay goose` and
+        // `pay --alt goose` select a paid provider and use the payer proxy.
+        let alternate = prepare_alternate_provider(
+            AlternateClient::Goose,
+            &self.args,
+            network_override,
+            active_account_name,
+        )?;
+        let mut args = vec!["session".to_string(), "--with-extension".to_string()];
+        args.push(pay_mcp_command(pay_bin));
+        args.extend(args_without_model(&self.args));
+        launch_goose(&args, Some(&alternate), active_account_name)
+    }
+}
+
+fn launch_goose(
+    args: &[String],
+    alternate: Option<&AlternateProvider>,
+    active_account_name: Option<&str>,
+) -> pay_core::Result<i32> {
+    let mut command = Command::new("goose");
+    command
+        .args(args)
+        .stdin(Stdio::inherit())
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit());
+
+    if let Some(account) = active_account_name {
+        command.env("PAY_ACTIVE_ACCOUNT", account);
+    }
+    if let Some(alternate) = alternate {
+        let model = alternate.model.as_deref().ok_or_else(|| {
+            pay_core::Error::Config(
+                "Goose alternate-provider routing requires a selected model".to_string(),
+            )
+        })?;
+        command.envs(goose_provider_env(alternate, model));
+    }
+
+    let status = command.status().map_err(|e| {
+        pay_core::Error::Config(format!(
+            "Failed to launch Goose: {e}. Install: https://block.github.io/goose/docs/getting-started/installation"
+        ))
+    })?;
+    Ok(status.code().unwrap_or(1))
+}
+
+fn goose_provider_env(alternate: &AlternateProvider, model: &str) -> Vec<(String, String)> {
+    vec![
+        (GOOSE_PROVIDER_ENV.to_string(), "openai".to_string()),
+        (GOOSE_MODEL_ENV.to_string(), model.to_string()),
+        (OPENAI_HOST_ENV.to_string(), alternate.base_url.clone()),
+        (OPENAI_API_KEY_ENV.to_string(), "pay".to_string()),
+        (
+            OPENAI_BASE_PATH_ENV.to_string(),
+            "v1/chat/completions".to_string(),
+        ),
+        // Goose otherwise makes a second, hidden model request after each of
+        // the first few turns to generate a session title. With a paid
+        // provider that request opens and settles its own payment channel.
+        (
+            GOOSE_DISABLE_SESSION_NAMING_ENV.to_string(),
+            "true".to_string(),
+        ),
+    ]
+}
+
+fn goose_version_requested(args: &[String]) -> bool {
+    args.iter()
+        .any(|arg| matches!(arg.as_str(), "--version" | "-V"))
+}
+
+fn goose_help_requested(args: &[String]) -> bool {
+    args.iter()
+        .any(|arg| matches!(arg.as_str(), "--help" | "-h"))
+}
+
+fn args_without_model(args: &[String]) -> Vec<String> {
+    let mut out = Vec::with_capacity(args.len());
+    let mut iter = args.iter();
+    while let Some(arg) = iter.next() {
+        if matches!(arg.as_str(), "--model" | "-m") {
+            let _ = iter.next();
+            continue;
+        }
+        if arg.starts_with("--model=") {
+            continue;
+        }
+        out.push(arg.clone());
+    }
+    out
+}
+
+fn pay_mcp_command(pay_bin: &str) -> String {
+    #[cfg(windows)]
+    return format!("\"{}\" mcp", pay_bin.replace('"', "\"\""));
+
+    #[cfg(not(windows))]
+    format!("'{}' mcp", pay_bin.replace('\'', "'\\''"))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn alternate_provider_uses_goose_openai_environment() {
+        let provider = AlternateProvider {
+            base_url: "http://127.0.0.1:54321".to_string(),
+            model: Some("qwen3.7-plus".to_string()),
+        };
+        let env = goose_provider_env(&provider, "qwen3.7-plus");
+
+        assert!(env.contains(&(GOOSE_PROVIDER_ENV.to_string(), "openai".to_string())));
+        assert!(env.contains(&(
+            OPENAI_HOST_ENV.to_string(),
+            "http://127.0.0.1:54321".to_string()
+        )));
+        assert!(env.contains(&(
+            OPENAI_BASE_PATH_ENV.to_string(),
+            "v1/chat/completions".to_string()
+        )));
+        assert!(env.contains(&(
+            GOOSE_DISABLE_SESSION_NAMING_ENV.to_string(),
+            "true".to_string()
+        )));
+    }
+
+    #[test]
+    fn selected_model_is_not_forwarded_to_goose_session() {
+        assert_eq!(
+            args_without_model(&["--model".into(), "qwen3.7-plus".into(), "--debug".into()]),
+            vec!["--debug"]
+        );
+    }
+
+    #[test]
+    fn metadata_requests_do_not_start_alternate_routing() {
+        assert!(goose_version_requested(&["--version".into()]));
+        assert!(goose_help_requested(&["--help".into()]));
+        assert!(!goose_version_requested(&["hello".into()]));
+        assert!(!goose_help_requested(&["hello".into()]));
+    }
+
+    #[test]
+    fn mcp_command_quotes_spaces() {
+        let command = pay_mcp_command("/tmp/Pay Tools/pay");
+        assert!(command.contains("Pay Tools"));
+        assert!(command.ends_with(" mcp"));
+    }
+}
--- a/rust/crates/cli/src/commands/goose.rs
+++ b/rust/crates/cli/src/commands/goose.rs
@@ -5,6 +5,7 @@ use std::process::{Command, Stdio};
 
 use clap::Args;
 
+use super::agent_args::args_without_model;
 use super::claude::{AlternateClient, AlternateProvider, prepare_alternate_provider};
 
 const GOOSE_PROVIDER_ENV: &str = "GOOSE_PROVIDER";
@@ -117,22 +118,6 @@ fn goose_help_requested(args: &[String]) -> bool {
         .any(|arg| matches!(arg.as_str(), "--help" | "-h"))
 }
 
-fn args_without_model(args: &[String]) -> Vec<String> {
-    let mut out = Vec::with_capacity(args.len());
-    let mut iter = args.iter();
-    while let Some(arg) = iter.next() {
-        if matches!(arg.as_str(), "--model" | "-m") {
-            let _ = iter.next();
-            continue;
-        }
-        if arg.starts_with("--model=") {
-            continue;
-        }
-        out.push(arg.clone());
-    }
-    out
-}
-
 fn pay_mcp_command(pay_bin: &str) -> String {
     #[cfg(windows)]
     return format!("\"{}\" mcp", pay_bin.replace('"', "\"\""));
@@ -166,14 +151,6 @@ mod tests {
             GOOSE_DISABLE_SESSION_NAMING_ENV.to_string(),
             "true".to_string()
         )));
-    }
-
-    #[test]
-    fn selected_model_is_not_forwarded_to_goose_session() {
-        assert_eq!(
-            args_without_model(&["--model".into(), "qwen3.7-plus".into(), "--debug".into()]),
-            vec!["--debug"]
-        );
     }
 
     #[test]
--- a/rust/crates/cli/src/commands/payer_proxy.rs
+++ b/rust/crates/cli/src/commands/payer_proxy.rs
@@ -8,13 +8,11 @@
 //! 1. Every request is forwarded upstream, preserving method, path+query,
 //!    and headers (minus hop-by-hop). The request body is buffered (capped
 //!    at [`MAX_BODY_BYTES`]) so it can be replayed.
-//! 2. When the upstream answers `402 Payment Required` with an MPP
-//!    `Payment` challenge, the proxy builds a signed charge credential
-//!    with the exact client machinery that backs `pay curl` / `pay fetch`
-//!    ([`pay_core::mpp::select_challenge_by_balance`] +
-//!    [`pay_core::mpp::build_credential`]), swaps the request's
-//!    `Authorization` header for `Payment <credential>`, and retries the
-//!    buffered request exactly once.
+//! 2. When the upstream answers `402 Payment Required`, the proxy satisfies
+//!    the configured payment protocol and retries the buffered request once.
+//!    Hosted inference routes can require an MPP delegated session; the
+//!    resulting channel authorization is cached for the lifetime of the agent
+//!    process so subsequent completions do not create new on-chain payments.
 //! 3. Response bodies stream back ([`Body::from_stream`]).
 //! 4. If payment fails for any reason (mainnet challenge under
 //!    `--sandbox`, insufficient funds, signer errors, …) the original 402
@@ -35,6 +33,7 @@
 //! challenge fires on the translated request, so the retry replays the
 //! translated body. Other requests and dialects pass through untouched.
 
+use std::net::{Ipv4Addr, SocketAddr};
 use std::sync::Arc;
 
 use axum::Router;
@@ -58,6 +57,10 @@ use crate::commands::server::inference::providers::Dialect;
 /// buffer so a runaway client cannot exhaust memory.
 const MAX_BODY_BYTES: usize = 32 * 1024 * 1024;
 
+/// The payer proxy is an implementation detail shared only with the child
+/// agent process. Never bind it to a LAN-reachable interface.
+const PAYER_PROXY_BIND_IP: Ipv4Addr = Ipv4Addr::LOCALHOST;
+
 /// Handle returned by [`start_background`].
 pub struct PayerProxy {
     /// Base URL of the payer proxy, e.g. `http://127.0.0.1:52341`.
@@ -66,6 +69,15 @@ pub struct PayerProxy {
     /// startup (in sandbox mode the ephemeral localnet wallet is created
     /// eagerly so the pubkey is always known).
     pub payer_pubkey: Option<String>,
+}
+
+/// Payment contract the local payer must enforce for its upstream.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub enum PaymentProtocol {
+    /// Preserve the generic payer behavior: MPP charge first, then x402 upto.
+    Auto,
+    /// Require a delegated MPP session and never fall back to x402.
+    MppSession,
 }
 
 /// Where the payer forwards to, and how to talk to it.
@@ -86,6 +98,15 @@ pub struct PayerUpstream {
     /// translated requests are sent to, e.g. `v1/chat/completions` or
     /// Alibaba's `compatible-mode/v1/chat/completions`.
     pub chat_path: String,
+    /// Responses API path used by Codex, e.g.
+    /// `compatible-mode/v1/responses` for Alibaba Model Studio.
+    pub responses_path: String,
+    /// Refuse successful responses that were served without a 402 handshake.
+    /// Hosted catalog providers set this so a gateway routing mistake cannot
+    /// silently bypass metering and consume upstream credentials for free.
+    pub require_payment: bool,
+    /// Payment protocol required by this route.
+    pub payment_protocol: PaymentProtocol,
 }
 
 /// Shared state for the proxy handlers.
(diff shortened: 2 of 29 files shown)


web-ui/proxy/alibaba-qwen.yml
```diff
@@ -0,0 +1,201 @@
...
+    label: "Gateway"
+
+# Streaming agent sessions are metered from the usage reported by Model Studio.
+session:
+  cap_usdc: 10.0
+  min_voucher_delta: 1
+  close_delay_ms: 15000
+  settlement_interval_ms: 5000
+  modes: [pull]
+  pull_voucher_strategy: client_voucher
```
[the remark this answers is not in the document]
lgalabru: Fixed in 70c21f1: both hosted agent specs now explicitly set settlement_authority: delegated, matching the payer proxy requirement.

rust/crates/cli/src/commands/server/inference/providers/catalog.rs
```diff
@@ -16,6 +16,13 @@ pub const DEFAULT_CATALOG_FQNS: &[&str] = &[
     "solana-foundation/google/generativelanguage",
 ];
 
+const ALIBABA_MODELSTUDIO_FQN: &str = "solana-foundation/alibaba/modelstudio";
+const ALIBABA_MODELSTUDIO_GATEWAY_URL: &str = "https://modelstudio.alibaba.gateway-402.com";
+const ALIBABA_RESPONSES_PATH: &str = "compatible-mode/v1/responses";
```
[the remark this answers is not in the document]
lgalabru: Fixed in 70c21f1: Alibaba Responses now stays on v1/responses in the catalog and payer proxy, matching the metered gateway route. The route-contract tests were updated and pass.

rust/crates/cli/src/commands/payer_proxy.rs
```diff
@@ -552,6 +754,72 @@ impl PaidHeaders {
...
+    if request.settlement_authority != SessionSettlementAuthority::Delegated {
+        return Err(pay_core::Error::Mpp(
+            "agent payer requires a delegated MPP session".to_string(),
+        ));
+    }
+    if !request.modes.is_empty() && !request.modes.contains(&SessionMode::Push) {
+        return Err(pay_core::Error::Mpp(
+            "agent payer requires MPP session push mode".to_string(),
+        ));
+    }
```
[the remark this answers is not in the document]
lgalabru: Fixed in 40a91a6. Both hosted agent specs now advertise delegated push sessions and no longer set a client-voucher pull strategy. This matches the payer proxy contract: the client opens one payment channel, reuses its open authorization, and the gateway operator signs metered cumulative vouchers.

web-ui/proxy/google-gemini.yml
```diff
@@ -207,6 +209,17 @@ endpoints:
...
+  # Agent harnesses such as Claude, Goose, and Qoder speak OpenAI Chat
+  # Completions. Google exposes the same Gemini models through this official
+  # compatibility surface, so they do not need a Gemini-native client.
+  - method: POST
+    path: "v1beta/openai/chat/completions"
+    resource: "chat"
+    description: "OpenAI-compatible chat completions for Gemini models."
+    metering:
+      schemes: [mpp-session]
+      variants: *official_gemini_generation_metering
```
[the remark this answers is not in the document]
lgalabru: Fixed in dc612a9. Buffered settlement now converts OpenAI-compatible `usage.prompt_tokens` / `usage.completion_tokens` into configured quota units, including inference-observer counts; SSE settlement parses the same usage object. Added focused 8-input/5-output regressions that produce 2 input and 3 output quota units.

rust/crates/core/src/server/metering.rs
```diff
@@ -402,20 +403,79 @@ fn extract_and_price_usage(
...
+    match dim.unit {
+        BillingUnit::Tokens => Some(tokens),
+        BillingUnit::QuotaUnits => {
+            quota_tokens_per_unit(dim).map(|per_unit| ceil_div_u64(tokens, per_unit))
+        }
+        _ => None,
+    }
+}
+
+fn quota_tokens_per_unit(dim: &MeterDimension) -> Option<u64> {
```
EfeDurmaz16: blocking: `quota_tokens_per_unit` reads the first tier's `tokens per quota unit` hint and ignores each tier's `condition`. Both gemini-2.5-pro input tiers carry the same `price_usd: 0.000005`; the >200k premium lives only in the hint (4 vs 2 tokens/unit) and matches Google's real rates ($1.25/1M at `context_length <= 200000`, $2.50/1M above, https://ai.google.dev/gemini-api/docs/pricing). The code always uses 4, so a >200k request settles $1.25/1M instead of $2.50/1M, a 50% undercharge. Same defect in `SessionUsageHints::from_metering` (session_stream.rs:98), the mpp-session path these endpoints declare. Fix: pick the hint from the tier `resolve_tier` selects, or move the premium into `price_usd`.

web-ui/proxy/alibaba-qwen.yml
```diff
@@ -0,0 +1,201 @@
...
+          - price_usd: 2.294
+            condition: { field: context_length, op: "<=", value: 32000 }
+          - price_usd: 3.441
+            condition: { field: context_length, op: "<=", value: 128000 }
+          - price_usd: 5.735
+            condition: { field: context_length, op: "<=", value: 256000 }
+          - price_usd: 28.671
+            condition: { field: context_length, op: ">", value: 256000 }
+  - param: model
+    value: default
```
EfeDurmaz16, line 151: blocking: this `default` variant is never selected server side. `resolve_variant` (metering.rs:694) is `hint.contains(&v.value)` with a `variants.first()` fallback (same in `session_metering.rs` and `session_stream.rs`); no server resolver treats `value: default` as a catch-all, and `test_resolve_price` (metering.rs:1351) already encodes unmatched to first. An uncatalogued DashScope model (qwen-max, ...) bills at qwen3.7-plus (~$1-3/1M) instead of ~$28/1M. The client `catalog.rs match_variant` does honor `default`, so CLI estimates diverge from settlement. Fix: honor `default` in the shared server resolver, or reject uncatalogued models. `google-gemini.yml` has no `default` and its first variant is the priciest, so the same fallback overcharges there.

rust/crates/core/src/server/metering.rs
```diff
@@ -402,20 +403,79 @@ fn extract_and_price_usage(
...
+                let thoughts = usage
+                    .get("thoughtsTokenCount")
+                    .and_then(serde_json::Value::as_u64)
+                    .unwrap_or(0);
+                let derived = usage
+                    .get("totalTokenCount")
+                    .and_then(serde_json::Value::as_u64)
+                    .and_then(|total| total.checked_sub(input))
+                    .unwrap_or(0);
+                Some(candidates.saturating_add(thoughts).max(derived))
```
EfeDurmaz16, line 504: important: this returns `max(candidates+thoughts, total-prompt)`, and Gemini's `totalTokenCount` includes `toolUsePromptTokenCount` (billed at the input rate). On tool-use or grounding responses `derived` wins and those input tokens are priced as output quota units. Overcharge, capped by the signed ceiling. Use `derived - toolUsePromptTokenCount`.

rust/crates/cli/src/commands/payer_proxy.rs
```diff
@@ -363,17 +524,58 @@ async fn proxy(State(state): State<Arc<PayerState>>, req: Request) -> Response {
...
-        Ok(retry) => deliver(retry, translated).await,
+        Ok(retry) => {
+            if retry.status() == StatusCode::PAYMENT_REQUIRED
+                && let Some(cache) = session_authorization.as_mut()
+            {
+                **cache = None;
+            }
+            deliver(retry, translated).await
+        }
         Err(e) => {
```
EfeDurmaz16, line 549: important: a transport error on the paid retry clears the cached idempotent open credential, but the push-mode open may already be funded on chain (server/session.rs:1043). The next request consumes a fresh challenge with a fresh salt and opens a second funded channel: double open, orphaned escrow. Keep the credential on transport errors and clear only on a definitive server reject.

rust/crates/core/src/server/session.rs
```diff
@@ -919,11 +1137,17 @@ impl SessionMpp {
...
 
             SessionAction::Close(p) => {
-                let params = self
-                    .server
-                    .process_close(p)
-                    .await
-                    .map_err(|e| Error::Mpp(format!("Session close failed: {e}")))?;
+                let params = match self.server.process_close(p).await {
+                    Ok(params) => params,
+                    Err(error) if session_close_needs_reconciliation(&error) => self
```
EfeDurmaz16, line 1164: important: the client-facing Close reconciliation arm has no `reserve_capacity`/lock (only the delegated-voucher path at 884 is locked). A client close racing the runloop auto-close both reach `submit_payment_channel_settlement` and, in the pre-confirmation window, build the same-watermark settle+seal into the dedup-free batch: duplicate broadcasts, wasted operator SOL. Guard it with the runloop's reservation.

rust/crates/cli/src/commands/goose.rs
```diff
@@ -0,0 +1,193 @@
...
+    args.iter()
+        .any(|arg| matches!(arg.as_str(), "--version" | "-V"))
+}
+
+fn goose_help_requested(args: &[String]) -> bool {
+    args.iter()
+        .any(|arg| matches!(arg.as_str(), "--help" | "-h"))
+}
+
+fn args_without_model(args: &[String]) -> Vec<String> {
```
EfeDurmaz16: suggestion: `args_without_model` here is byte-identical to `qodercli/mod.rs:225` and re-implements the `--model/-m/--model=` grammar that `claude/mod.rs:726 model_arg` already owns. Fold into one shared helper so the billed-model lock stays in sync.

rust/crates/core/src/server/metering.rs
```diff
@@ -429,10 +489,38 @@ fn extract_dimension_quantity(
     {
         return Ok(quantity);
     }
+    // Responses and Chat Completions use different names for the same token
+    // dimensions. Let the endpoint-level Responses preset override meters
+    // inherited from a shared model-pricing table.
+    if preset.is_some_and(|preset| preset.eq_ignore_ascii_case("openai-responses"))
```
EfeDurmaz16, line 539: suggestion: this early-return fires for any token dim under `usage_preset=openai-responses` before the `dim.meter` branch, so a `ResponseHeader` meter is ignored and body dependence is forced. Ordering the `dim.meter` branch first keeps header-metered dims cheap.

EfeDurmaz16 (commented):
The close-watermark design is sound: skipping the voucher once the chain holds the accepted watermark is the right fix for the 0xEA duplicate settle. Two new pricing paths settle wrong money and a few money-adjacent races need a look, all inline.

rust/crates/core/src/server/metering.rs
lgalabru: Fixed in 5b4fbbd. Quota conversion now reads the notes from the same condition-selected tier used for pricing, in both buffered settlement and SessionUsageHints. Added >200k regressions proving the 2-token quota hint is selected.

web-ui/proxy/alibaba-qwen.yml
lgalabru: Fixed in 5b4fbbd. The shared server resolver now treats value: default as the fallback after trying named variants, and all buffered/session/streaming paths use it. Legacy first-variant fallback remains only for specs without an explicit default or top-level dimensions.

rust/crates/core/src/server/metering.rs
lgalabru: Fixed in 5b4fbbd. Gemini derived output now subtracts toolUsePromptTokenCount before comparison with candidates+thoughts in both buffered and SSE metering. Added regressions for both paths.

rust/crates/cli/src/commands/payer_proxy.rs
lgalabru: Fixed in 5e08e2d. Ambiguous paid-retry transport failures now preserve the idempotent open credential; only a definitive 402 clears it. A raw-socket regression closes the upstream after the challenge and verifies the cached credential survives.

rust/crates/core/src/server/session.rs
lgalabru: Fixed in 5b4fbbd. Client Close now owns the same per-channel capacity lease as the runloop/delegated path through reconciliation and settlement, preventing duplicate close broadcasts. The close-flow test now exercises and rejects a competing reservation.

rust/crates/cli/src/commands/goose.rs
lgalabru: Addressed in e949957. The --model/-m/--model= parser and stripping logic now live in one shared agent_args module used by Claude, Goose, and Qoder, with centralized grammar tests.

rust/crates/core/src/server/metering.rs
lgalabru: Addressed in 5b4fbbd. Explicit ResponseHeader meters now run before the OpenAI Responses preset and are excluded from response-body requirements. The preset still overrides inherited ResponseJson paths; a no-body header-meter regression covers the distinction.

lgalabru:
Addressed the remaining 4/5 streaming payment-gate finding in signed commit 8edcf32. `meter_response_stream` now waits for the required cumulative commit before yielding the chargeable chunk, rather than releasing the chunk first. Added an end-to-end regression that proves output stays withheld while the watermark is pending and resumes only after the matching voucher is accepted; all 8 session-stream tests pass.

rust/crates/cli/src/commands/payer_proxy.rs
```diff
@@ -283,10 +325,38 @@ async fn proxy(State(state): State<Arc<PayerState>>, req: Request) -> Response {
...
 
-    let first = match send_upstream(&state, &method, &url, &headers, body.clone(), None).await {
+    // A delegated session is shared across every request made by this agent
+    // process. Keep the guard until the gateway has accepted and metered this
+    // response so two concurrent calls cannot reserve the same channel cap.
+    let mut session_authorization = if state.payment_protocol == PaymentProtocol::MppSession {
+        Some(state.session_authorization.lock().await)
+    } else {
+        None
+    };
```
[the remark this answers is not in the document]
lgalabru: Fixed in 36fdac6. The cached authorization mutex is now acquired as an owned guard and moved into passthrough and translated streaming bodies, so it is released only at EOF or body drop. Added a concurrent slow-stream regression proving a second request cannot reach upstream until the first body completes.

web-ui/proxy/alibaba-qwen.yml
```diff
@@ -0,0 +1,201 @@
...
+
+endpoints:
+  # Claude Code and other OpenAI-chat clients are routed through this surface.
+  - method: POST
+    # The deployed Agent Gateway exposes Model Studio's OpenAI surface under
+    # its upstream-compatible prefix. Agent harnesses still use the standard
+    # `/v1/chat/completions`; the payer proxy maps that path to this one.
+    path: "compatible-mode/v1/chat/completions"
+    resource: "chat"
+    description: "OpenAI-compatible streaming chat completions for Qwen models."
```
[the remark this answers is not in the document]
lgalabru: Fixed in 7724d3f. Delegated SSE responses now stay streaming: each observable chunk is rated and the gateway-signed cumulative voucher is persisted before that chunk is released. The capacity lease remains owned by the response body through EOF/drop, so large streams no longer go through the 1 MiB buffer or turn into a post-inference 502. Added OpenAI/Gemini live-usage observation plus a paused-upstream regression proving chunk one reaches the client before EOF with its matching watermark already committed.

web-ui/proxy/google-gemini.yml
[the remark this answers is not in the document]
lgalabru: Fixed in 7724d3f. Delegated SSE responses now stay streaming: each observable chunk is rated and the gateway-signed cumulative voucher is persisted before that chunk is released. The capacity lease remains owned by the response body through EOF/drop, so large streams no longer go through the 1 MiB buffer or turn into a post-inference 502. Added OpenAI/Gemini live-usage observation plus a paused-upstream regression proving chunk one reaches the client before EOF with its matching watermark already committed.

rust/crates/core/src/server/gate.rs
```diff
@@ -1552,22 +1572,22 @@ async fn session_authorized(
             };
             let variant = variant_hint_from_path(path);
             let ceiling_usd = available_base_units as f64 / 10_f64.powi(sm.decimals() as i32);
-            let settlement = Some(Box::new(metering::UptoSettlementPlan {
+            let settlement = metering::UptoSettlementPlan {
                 metering: meter.clone(),
                 variant_hint: variant,
```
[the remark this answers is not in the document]
lgalabru: Fixed in 0e19f6c. When a delegated settlement plan has no path-derived variant, the payment middleware now reads the root `model` from the OpenAI/Anthropic-compatible JSON request, restores the exact body for the upstream handler, and applies that model to the settlement plan. Native `models/{id}` routes keep their path hint. Added focused coverage for selected, missing, blank, and malformed model fields.

rust/crates/core/src/server/payment.rs
```diff
@@ -92,11 +94,25 @@ async fn gate_adapter<S: PaymentState>(state: S, req: Request<Body>, next: Next)
...
+                        .is_some_and(|plan| plan.variant_hint.is_none())
+                    {
+                        let (restored, body_variant) = match request_body_variant_hint(req).await {
+                            Ok(result) => result,
+                            Err(response) => return response,
+                        };
+                        req = restored;
+                        if let Some(plan) = sf.settlement.as_deref_mut() {
+                            plan.variant_hint = body_variant;
+                        }
```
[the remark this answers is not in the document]
lgalabru: Fixed in ff44b8c. Delegated request parsing now derives a tokenizer-independent context-token estimate from the actual prompt-bearing JSON fields (`messages`, `input`, `prompt`, `instructions`, `system`, `contents`, and `tools`) and applies it to `RequestProperties.context_length` before constructing the stream meter. The estimator treats non-ASCII conservatively, preserves path-derived variants, and restores the exact upstream body. Added coverage proving a long Gemini request crosses the configured 200k high-context tier boundary.

rust/crates/core/src/server/payment.rs
```diff
@@ -192,6 +213,94 @@ async fn gate_adapter<S: PaymentState>(state: S, req: Request<Body>, next: Next)
...
+/// tier before provider usage arrives. ASCII text averages roughly three or
+/// more bytes per token for agent prompts; non-ASCII scalars count as one each.
+fn estimate_json_tokens(value: &serde_json::Value) -> u64 {
+    match value {
+        serde_json::Value::Null => 0,
+        serde_json::Value::Bool(_) | serde_json::Value::Number(_) => 1,
+        serde_json::Value::String(text) => {
+            let ascii = text.chars().filter(char::is_ascii).count() as u64;
+            let non_ascii = text.chars().filter(|ch| !ch.is_ascii()).count() as u64;
+            ascii.saturating_add(2) / 3 + non_ascii
```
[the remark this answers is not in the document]
lgalabru: Fixed in e1fe478. Removed the tokenizer-independent character heuristic entirely. Buffered responses now select context_length tiers from the provider reported prompt token count (OpenAI prompt_tokens or Gemini promptTokenCount). Delegated SSE streams re-rate their cumulative meter when the provider usage event arrives, preserve the prior committed watermark, and persist any higher-tier delta before releasing that event. Added regressions for both buffered tier selection and streamed cumulative re-rating; all 753 pay-core server tests pass.

rust/crates/core/src/server/session_stream.rs
```diff
@@ -366,6 +562,17 @@ impl StreamUsageAccumulator {
                     changed = true;
                 }
             }
+
+            if let Some(usage) = value.get("usage") {
```
EfeDurmaz16: blocking: this only reads `usage.prompt_tokens` / `usage.completion_tokens` (plus `usageMetadata` above), but the two endpoints this PR ships in `alibaba-qwen.yml` report neither. OpenAI Responses puts usage under `/response/usage` as `input_tokens` / `output_tokens` on `response.completed`, and Anthropic reports `message.usage.input_tokens` on `message_start` plus a top-level `usage.output_tokens` on `message_delta` (https://platform.claude.com/docs/en/api/messages-streaming), so `output_tokens` stays 0 and Output/Tokens settles at the whitespace word floor while input settles at 0.

The non-streaming path is worse: `quota_token_quantity` (metering.rs:450) misses the same fields, extraction returns None, and the default `MissingUsagePolicy::Refund` settles zero after the body is delivered. `proxy/src/observer.rs:191-215` already parses all three shapes, so reusing it here is the small fix. Do not gate the stream path back to buffering, that restores the zero settle.

web-ui/proxy/alibaba-qwen.yml
```diff
@@ -0,0 +1,201 @@
...
+      type: proxy
+      url: https://dashscope-us.aliyuncs.com
+      auth: *dashscope_auth
+    metering:
+      schemes: [mpp-session]
+      variants: *qwen_model_pricing
+
+  # Codex uses the OpenAI Responses API wire format.
+  - method: POST
+    path: "v1/responses"
```
EfeDurmaz16, line 182: blocking (same root): this endpoint and `v1/messages` below declare `unit: tokens` dimensions with no `usage_preset` and no explicit meter pointers, so they fall to the generic parser that only knows the OpenAI chat and Gemini shapes. Declaring the Responses and Anthropic usage pointers here fixes it without touching the metering code.

rust/crates/core/src/server/payment.rs
```diff
@@ -202,6 +248,39 @@ async fn settle_axum_delegated_response(
...
+    let is_sse = response
+        .headers()
+        .get(header::CONTENT_TYPE)
+        .and_then(|value| value.to_str().ok())
+        .is_some_and(|value| {
+            value
+                .split(';')
+                .any(|part| part.trim().eq_ignore_ascii_case("text/event-stream"))
+        });
+    if is_sse && session_stream::DelegatedSessionStreamMeter::supports(&forward) {
```
EfeDurmaz16, line 295: important: delegated SSE only sees prompt tokens if the provider volunteers a usage frame, which for OpenAI-compatible upstreams needs `stream_options.include_usage` in the request, and only the Claude translation path adds it (claude/translate.rs:57). Every other streamed session bills input at 0.

rust/crates/core/src/server/session_stream.rs
```diff
@@ -147,6 +148,163 @@ pub struct SessionStreamMeter {
...
+        let committed = self.gate.committed_base_units();
+        let amount = target.saturating_sub(committed);
+        if amount == 0 {
+            return Ok(());
+        }
+        let authorized = self
+            .forward
+            .committed_base_units
+            .saturating_add(self.forward.available_base_units);
+        if target > authorized {
```
EfeDurmaz16, line 273: important: `settle()` returns Err when `target > authorized`, which is right for additive growth since the chunk is never yielded, but `reprice_for_context_length` retroactively re-rates output that was already yielded and committed. In that case the whole settle aborts instead of clamping to `authorized`, so the operator collects nothing for output the client already has.

rust/crates/cli/src/commands/payer_proxy.rs
```diff
@@ -703,23 +982,38 @@ async fn send_upstream(
...
-        .body(Body::from_stream(resp.bytes_stream()))
-        .unwrap_or_else(|_| {
-            (
-                StatusCode::INTERNAL_SERVER_ERROR,
-                "payer proxy: response build error",
-            )
-                .into_response()
-        })
+    let stream_body = Body::from_stream(async_stream::stream! {
+        let _session_guard = session_guard;
```
EfeDurmaz16, line 1002: important: moving the owned session guard into the response body is right for metering, but it now holds the process-wide session mutex for the whole upstream read, and the client at line 156 is built with no `timeout`, `read_timeout` or `connect_timeout`. One hung provider stream blocks every other paid request on that session indefinitely.

rust/crates/core/src/server/session.rs
```diff
@@ -527,8 +725,21 @@ impl SessionMpp {
...
+            challenge_binding_secret,
+            Arc::new(MemoryChannelStore::new()),
+        )
+    }
+
+    /// Create with a caller-provided durable channel store.
+    pub fn new_with_channel_store(
+        config: SessionConfig,
+        challenge_binding_secret: impl Into<String>,
+        channel_store: Arc<dyn ChannelStore>,
```
EfeDurmaz16, line 740: important: the channel store is now pluggable and Redis-backed, but `committed_watermarks` stays a process-local `HashMap` (session.rs:78), so after a restart `authorize_delegated_usage` fails with "unknown delegated session channel" for a channel Redis still holds. It fails closed rather than losing money, but it defeats the durability this PR advertises; seeding the watermark from the store on a miss would close it.

EfeDurmaz16 (commented):
Re-reviewed the 9 new commits. All six prior findings are fixed and the tier selector, default variant, close lock and credential retention now have regression tests behind them. One new blocker: the Alibaba Responses and Anthropic endpoints this PR ships are metered by a parser that cannot read their usage dialect, so they settle at a word count or at zero. Details inline.

rust/crates/core/src/server/session_stream.rs
lgalabru: Valid distinction, fixed in ffc2a20. Earlier low-tier vouchers meant collection was not literally zero, but the remaining authorization could be stranded. A usage-only retroactive reprice now commits up to the authorized cap, marks the reservation exhausted so finalization completes cleanly, and still fails closed for subsequent/additive visible output. The regression exercises a 3 -> 15 reprice with a cap of 10.

rust/crates/cli/src/commands/payer_proxy.rs
lgalabru: Valid and fixed in ffc2a20. The payer client now has a 15-second connect timeout and a 300-second idle read timeout, while deliberately retaining no total timeout so healthy long-lived completion streams continue working. A silent provider can no longer hold the session mutex forever.

rust/crates/core/src/server/session.rs
lgalabru: Valid and fixed in ffc2a20. The durable `ChannelStore` is now authoritative on every delegated authorization; its cumulative watermark repopulates the process cache and allows a restarted or different replica to continue safely. The regression clears the local map after cumulative 75 and verifies the next authorization resumes at 100.

web-ui/proxy/alibaba-qwen.yml
lgalabru: Same valid root cause, fixed centrally in ffc2a20. I intentionally did not add YAML response pointers: static pointers cannot represent Anthropic input/output usage split across multiple SSE events, and the stream meter does not consume response-json pointers. The shared parser now covers both these endpoints in buffered and streamed paths.

rust/crates/core/src/server/payment.rs
lgalabru: Valid and fixed in ffc2a20. Delegated streamed Chat Completions requests now force `stream_options.include_usage=true` before upstream forwarding and rewrite `Content-Length` after serialization. Native Gemini, Responses, and Anthropic surfaces already emit their usage events.

rust/crates/core/src/server/session_stream.rs
lgalabru: Valid and fixed in ffc2a20. `provider_token_quantity` is now the shared buffered/streaming parser for Chat Completions, nested Responses usage, Anthropic message/top-level usage, and Gemini usage metadata. Added SSE and buffered regressions for every shape; streaming remains unbuffered.

EfeDurmaz16 (commented):
Re-reviewed `ffc2a20`. All six findings are closed and each one landed the minimal version: `provider_token_quantity` now reads `usage`, `/response/usage` and `/message/usage` with both key namings and is shared by the stream and buffered paths, `include_usage` is injected only on `chat/completions` so the Anthropic and Responses bodies stay untouched (both emit usage by default), reprice clamps to `authorized` instead of aborting the settle, the proxy client got connect and read timeouts, and `authorize_delegated_usage` seeds the watermark from the channel store so a restarted or second replica continues instead of failing. The two new tests cover exactly the parser and the forced usage frame. Nothing further from me.

EfeDurmaz16 (approved):

PR #412, opened 2026-07-27
fix(pay): restore builds, auth gates, and telemetry

lgalabru:
## Summary

- restore Windows release builds, add a Windows CLI compile check, and update release versions
- preserve authentication gates for file-backed server accounts and describe session authorization using the funded stablecoin limit and provider origin
- emit reliable payment, collection, settlement, delivery, upstream, and channel lifecycle metrics with bounded protocol labels such as `x402/exact`, `x402/upto`, `mpp/charge`, and `mpp/session`
- move the session idle-close clock from horizontally scaled proxies into the durable `ChannelStore`, with one-minute rounded deadlines for worker-side batching
- pin the reviewed PayKit lifecycle store implementation from solana-foundation/pay-kit#256

## Test Plan

- `cargo fmt --check`
- `cargo check -p pay -p pay-proxy --all-targets`
- `cargo test -p pay-proxy`
- GitHub Actions: Windows Rust check, Rust lint/tests, and TypeScript lint/tests

PR #413, opened 2026-07-28
fix(cli): preserve auth gates and clarify session spend

lgalabru:
Resolve server operator accounts through the network-aware loader so file-backed accounts retain their authentication policy while ephemeral accounts remain ungated.\n\nDescribe session authorization with the funded stablecoin limit and provider origin, using trusted decimals and raw base units for unknown assets.


lgalabru:
Consolidated into #412 so the release, authentication/session, and payment telemetry fixes can be reviewed and merged together.

PR #416, opened 2026-07-31
fix(session): keep active channels reusable

lgalabru:
## Summary

- keep delegated session lifecycle deadlines alive while an upstream request is in flight
- decouple idle-close reconciliation from shorter settlement wakeups
- retry transient cached-session store conflicts without opening a replacement channel
- preserve cached sessions on non-terminal 402 responses
- default omitted `session.close_delay_ms` values to ten minutes
- pin PayKit to `dc3a36ee` (merged squash of solana-foundation/pay-kit#259: reworked MPP session API + idle-close root fix + old-writer-safe channel records) and cascade: fallible challenge build that fails the challenge instead of silently degrading, `with_blockhash_cache` pass-through, end-to-end queued-touch-discard regression test
- share metering's model-support policy with provider discovery: a `default` pricing-variant sentinel keeps the full advertised model list (runtime prices unlisted models via that sentinel); non-chat entries are dropped by capability (`supportedGenerationMethods`) instead of by pricing; providers without a sentinel intersect with named families as before
- repair the network-gated openapi smoke test, broken on main since #398 (never built by CI)

### Picker/metering policy note

`resolve_price` also falls back to top-level endpoint `dimensions` (e.g. Model Studio's conservative list-price backstop), so the runtime technically prices every model for those providers too. That fallback deliberately does **not** admit models to the picker: only the `default` sentinel is the catalog author's promise that unlisted models are first-class; the dimensions backstop is quote safety at a punitive rate, and hiding those models steers users toward correctly-priced families instead of silently billing at the backstop. Documented on `models_matching_configured_families`.

`pay-kit#259` has merged and `dc3a36ee` (`d829fa0`) is the final pin — no further re-point pending.

2026-08-01 incident follow-up (`367dd9f`): mirrors PayKit's new pre-binding row split ("session channel predates proof binding" instead of the generic mismatch for rows that predate — or were stripped of — the proof binding), and drops the extra echoed-challenge-id comparison PayKit's canonical check doesn't have. Deploy alongside agent-gateway#14 (settlement worker pin) and the four sibling gateways.

### GCP identity tokens for IAM-locked upstreams (`bc95dca`)

Folded in per pay-cli thread: `oauth2` auth gains an optional `audience` and the special `token_url: gcp_metadata_identity`, minting an audience-bound OIDC identity token from the metadata server's `/identity` endpoint (raw JWT; cache TTL from its own `exp` claim). This is what invoking IAM-protected Cloud Run services (scale-to-zero micro-agents) needs — a `gcp_metadata` access token doesn't work there. Fail-closed validation: identity mode requires `audience`; `audience` or `scopes` with the wrong `token_url` are spec errors with actionable messages. Token cache key now includes audience. No ADC fallback — user credentials cannot mint audience-bound identity tokens, and the error says so.

### Picker "unpriced" root cause: skills cache wiped by racing refreshes (`6f773a7`, `7d3e568`)

The recurring all-models-unpriced picker state had a second cause beyond the family-intersect bug fixed in `2a6b14f`: `clean_stale_caches` deleted every `skills-*.json` except the file its own process just wrote, so two concurrent catalog refreshes (there is one `pay mcp` per agent session) could delete each other's write and leave **zero** cache files. With no cached catalog, picker discovery silently fell back to the built-in `google_gemini_fallback`/`alibaba_modelstudio_fallback` providers, which deliberately carry no pricing — every model rendered "unpriced". Reproduced live (cache absent → 11 Gemini models, all unpriced) and verified fixed (same state → catalog fetched, 42 models, all priced).

- `6f773a7` (core): prune only cache files whose filename timestamp is strictly older than the kept write — racing refreshes converge on the newest file instead of an empty directory.
- `7d3e568` (cli): on cache miss, picker discovery fetches via `load_skills_for` (which rewrites the cache) before resorting to the priceless fallbacks — self-heals even against older installed binaries that still race.

### Review response (`6c9c1ea`..`adc8e4d`)

Fixes for @EfeDurmaz16's five blockers (inline replies posted on each comment):

- `6c9c1ea` — `SessionSpec` now has `deny_unknown_fields`; migrates the two live gateway configs (`alibaba-qwen.yml`, `google-gemini.yml`) off the removed `settlement_authority: delegated` key, which had been silently defaulting `voucher_signer` to `client` since this PR's own `9440d18` rename
- `bbb2ff0` — `persist_touch` now clamps the idle-close deadline to the channel's negotiated `idle_timeout_seconds`, not just the operator's `close_delay_ms`
- `1004c4b` — `enforce_session_cap` and the signing path now share one `resolve_session_deposit`, so the enforced cap can no longer diverge from what actually gets signed
- `b86e3ca` — payer proxy: five previously-unrecognized terminal session errors (pinned as shared constants in `pay_core::server::session::terminal_errors`) plus a cache-order fix so an unconfirmed credential is never adopted before its retry proves non-402

The five follow-ups are tracked as-is, not fixed in this round; each has an inline acknowledgment on the PR. Of note, the `operator_close_channel` settle-retry gap is the Rust twin of `pay-kit#259`'s close re-drive blocker and should port the same fix pattern.

## Test plan

At `adc8e4d`:

- `cargo test --workspace --all-features` — 26 targets, 1768 passed, 0 failed (includes the repaired live-network openapi smoke test)
- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-features --all-targets` — clean
- `cargo check -p pay` (default features) — clean

### DIFF
--- a/rust/crates/cli/src/commands/mod.rs
+++ b/rust/crates/cli/src/commands/mod.rs
@@ -587,7 +587,10 @@ fn handle_outcome(
             let req: Option<SessionRequest> = challenge.request.decode().ok();
             let cap_display = req
                 .as_ref()
-                .map(|request| display_token_amount(&request.cap, &request.currency))
+                .and_then(|request| {
+                    session_deposit_amount(request)
+                        .map(|deposit| display_token_amount(deposit, &request.currency))
+                })
                 .unwrap_or_else(|| "unknown".to_string());
 
             if auto_pay {
@@ -611,11 +614,13 @@ fn handle_outcome(
                     "status": 402,
                     "protocol": "mpp-session",
                     "challenge": {
-                        "cap": req.as_ref().map(|r| &r.cap),
+                        "suggested_deposit": req.as_ref().and_then(session_deposit_amount),
                         "cap_display": cap_display,
                         "currency": req.as_ref().map(|r| &r.currency),
-                        "network": req.as_ref().and_then(|r| r.network.as_deref()),
-                        "min_voucher_delta": req.as_ref().and_then(|r| r.min_voucher_delta.as_deref()),
+                        "network": req.as_ref().map(|r| &r.method_details.network),
+                        "min_voucher_delta": req
+                            .as_ref()
+                            .and_then(|r| r.method_details.min_voucher_delta.as_deref()),
                         "recipient": req.as_ref().map(|r| &r.recipient),
                     },
                     "resource": resource_url,
@@ -1164,6 +1169,15 @@ fn mpp_challenges_within_cap(
     ))
 }
 
+/// Deposit a session challenge asks the client to lock: the server's
+/// suggestion, falling back to its minimum.
+fn session_deposit_amount(request: &SessionRequest) -> Option<&str> {
+    request
+        .suggested_deposit
+        .as_deref()
+        .or(request.minimum_deposit.as_deref())
+}
+
 fn enforce_session_cap(
     request: Option<&SessionRequest>,
     payment_cap: Option<u64>,
@@ -1176,7 +1190,8 @@ fn enforce_session_cap(
             "session payment cap requires a decoded SessionRequest".to_string(),
         ));
     };
-    let required_micro = amount_as_stablecoin_micro(&request.cap, &request.currency)?;
+    let required = session_deposit_amount(request).unwrap_or("0");
+    let required_micro = amount_as_stablecoin_micro(required, &request.currency)?;
 
     if required_micro <= payment_cap {
         return Ok(());
@@ -1705,105 +1720,49 @@ fn pay_session_and_retry(
     sandbox: bool,
     verbose: bool,
 ) -> pay_core::Result<()> {
-    use pay_kit::mpp::{SessionMode, SessionPullVoucherStrategy};
-
     let is_json = no_dna::should_json(output_fmt);
     validate_tool_request_before_signing(tool)?;
 
-    // Deposit = min_voucher_delta * 1000, clamped to [1 USDC, cap].
-    let min_delta = req
-        .and_then(|r| r.min_voucher_delta.as_deref())
-        .and_then(|s| s.parse::<u64>().ok())
-        .unwrap_or(1_000);
-    let cap = req
-        .and_then(|r| r.cap.parse::<u64>().ok())
-        .unwrap_or(1_000_000);
-    let deposit = (min_delta * 1_000).max(1_000_000).min(cap);
-    let cap_display = req
-        .map(|request| display_token_amount(&request.cap, &request.currency))
-        .unwrap_or_else(|| format!("{cap} base units"));
-    let deposit_display = req
-        .map(|request| display_token_amount(&deposit.to_string(), &request.currency))
-        .unwrap_or_else(|| format!("{deposit} base units"));
+    let Some(request) = req else {
+        return Err(pay_core::Error::Mpp(
+            "session challenge did not decode into a SessionRequest".to_string(),
+        ));
+    };
+
+    // Deposit: server suggestion, floored by its minimum; fall back to 1 USDC.
+    let minimum = request
+        .minimum_deposit
+        .as_deref()
+        .and_then(|value| value.parse::<u64>().ok())
+        .unwrap_or(0);
+    let deposit = request
+        .suggested_deposit
+        .as_deref()
+        .and_then(|value| value.parse::<u64>().ok())
+        .unwrap_or(1_000_000)
+        .max(minimum)
+        .max(1);
+    let deposit_display = display_token_amount(&deposit.to_string(), &request.currency);
 
     if verbose && !is_json {
         crate::components::print_notice(
             crate::components::NoticeLevel::Success,
             "Authorizing MPP session",
-            &payment_authorization_notice_body(&cap_display, Some(&deposit_display)),
+            &payment_authorization_notice_body(&deposit_display, Some(&deposit_display)),
         );
     }
 
-    let supports_push = req
-        .map(|r| r.modes.is_empty() || r.modes.contains(&SessionMode::Push))
-        .unwrap_or(true);
-    let use_pull = req
-        .map(|r| {
-            r.modes.contains(&SessionMode::Pull)
-                && (!supports_push
-                    || matches!(
-                        r.pull_voucher_strategy.as_ref(),
-                        Some(SessionPullVoucherStrategy::ClientVoucher)
-                    ))
-        })
-        .unwrap_or(false);
-
-    let auth_header = if use_pull {
-        let Some(request) = req else {
-            return Err(pay_core::Error::Mpp(
-                "pull-mode session requires a decoded SessionRequest".to_string(),
-            ));
-        };
-
-        let store = pay_core::accounts::FileAccountsStore::default_path();
-        match request.pull_voucher_strategy.as_ref() {
-            Some(SessionPullVoucherStrategy::ClientVoucher) => {
-                let (_handle, header) =
-                    pay_core::session::open_payment_channel_session_header_with_mode(
-                        challenge,
-                        request,
-                        &store,
-                        network_override,
-                        account_override,
-                        deposit,
-                        SessionMode::Pull,
-                        resource_url,
-                        sandbox,
-                    )?;
-                header
-            }
-            Some(SessionPullVoucherStrategy::OperatedVoucher) => {
-                return Err(pay_core::Error::Mpp(
-                    "operated-voucher pull sessions are no longer supported; \
-                     use a client-voucher payment-channel session instead"
-                        .to_string(),
-                ));
-            }
-            None => {
-                return Err(pay_core::Error::Mpp(
-                    "pull-mode session challenge missing pullVoucherStrategy".to_string(),
-                ));
-            }
-        }
-    } else {
-        if let Some(request) = req {
-            let store = pay_core::accounts::FileAccountsStore::default_path();
-            let (_handle, header) = pay_core::session::open_payment_channel_session_header(
-                challenge,
-                request,
-                &store,
-                network_override,
-                account_override,
-                deposit,
-                resource_url,
-                sandbox,
-            )?;
-            header
-        } else {
-            let (_handle, header) = pay_core::session::open_session_header(challenge, deposit)?;
-            header
-        }
-    };
+    let store = pay_core::accounts::FileAccountsStore::default_path();
+    let (_handle, auth_header) = pay_core::session::open_payment_channel_session_header(
+        challenge,
+        request,
+        &store,
+        network_override,
+        account_override,
+        deposit,
+        resource_url,
+        sandbox,
+    )?;
 
     let receipt_network = network_override
         .map(str::to_string)
--- a/rust/crates/cli/src/commands/mod.rs
+++ b/rust/crates/cli/src/commands/mod.rs
@@ -1178,6 +1178,26 @@ fn session_deposit_amount(request: &SessionRequest) -> Option<&str> {
         .or(request.minimum_deposit.as_deref())
 }
 
+/// The exact base-unit deposit `pay_session_and_retry` will sign for. Shared
+/// with `enforce_session_cap` so the cap can never be checked against a
+/// different amount than the one that gets locked: server suggestion,
+/// floored by its minimum, falling back to 1 USDC when the challenge omits
+/// both.
+fn resolve_session_deposit(request: &SessionRequest) -> u64 {
+    let minimum = request
+        .minimum_deposit
+        .as_deref()
+        .and_then(|value| value.parse::<u64>().ok())
+        .unwrap_or(0);
+    request
+        .suggested_deposit
+        .as_deref()
+        .and_then(|value| value.parse::<u64>().ok())
+        .unwrap_or(1_000_000)
+        .max(minimum)
+        .max(1)
+}
+
 fn enforce_session_cap(
     request: Option<&SessionRequest>,
     payment_cap: Option<u64>,
@@ -1190,8 +1210,8 @@ fn enforce_session_cap(
             "session payment cap requires a decoded SessionRequest".to_string(),
         ));
     };
-    let required = session_deposit_amount(request).unwrap_or("0");
-    let required_micro = amount_as_stablecoin_micro(required, &request.currency)?;
+    let required = resolve_session_deposit(request).to_string();
+    let required_micro = amount_as_stablecoin_micro(&required, &request.currency)?;
 
     if required_micro <= payment_cap {
         return Ok(());
@@ -1729,19 +1749,7 @@ fn pay_session_and_retry(
         ));
     };
 
-    // Deposit: server suggestion, floored by its minimum; fall back to 1 USDC.
-    let minimum = request
-        .minimum_deposit
-        .as_deref()
-        .and_then(|value| value.parse::<u64>().ok())
-        .unwrap_or(0);
-    let deposit = request
-        .suggested_deposit
-        .as_deref()
-        .and_then(|value| value.parse::<u64>().ok())
-        .unwrap_or(1_000_000)
-        .max(minimum)
-        .max(1);
+    let deposit = resolve_session_deposit(request);
     let deposit_display = display_token_amount(&deposit.to_string(), &request.currency);
 
     if verbose && !is_json {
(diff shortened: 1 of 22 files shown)


rust/crates/core/src/server/session.rs
```diff
@@ -1232,13 +1248,36 @@ impl SessionMpp {
...
+                if let Err(error) = lifecycle
+                    .touch(heartbeat_channel_id.clone(), unix_millis())
+                    .await
+                {
+                    tracing::warn!(
+                        channel_id = heartbeat_channel_id,
+                        %error,
+                        "failed to heartbeat active delegated session"
+                    );
+                    break;
```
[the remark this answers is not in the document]
lgalabru: Fixed in 4355762. Transient lifecycle touch failures are now logged and retried on the next heartbeat interval instead of terminating the heartbeat task.

rust/crates/core/src/server/session.rs
```diff
@@ -454,10 +454,12 @@ fn close_voucher_required(onchain_settled: u64, latest_accepted: u64) -> bool {
 pub struct DelegatedCapacityLease {
     runtime: SessionOperatorRuntime,
     channel_id: String,
+    heartbeat: tokio::task::JoinHandle<()>,
 }
 
 impl Drop for DelegatedCapacityLease {
     fn drop(&mut self) {
+        self.heartbeat.abort();
```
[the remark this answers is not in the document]
lgalabru: Fixed in 4355762. Heartbeat touches now carry the lease cancellation signal; lease drop signals cancellation before aborting the task, and the lifecycle runloop uses a cancellation-biased select to discard queued or in-flight persistence. Added focused cancellation coverage; immediate capacity release and delegated heartbeat lifecycle tests still pass.

lgalabru:
Pushed three commits (`e221f78..2a6b14f`):

- **`e52e441` test(core)**: repair the `network_tests`-gated openapi smoke test — broken on main since #398 (`OpenapiSource::Path` became spec-dir-anchored, endpoint fields moved under `spec`); invisible to CI, surfaced by the full `--all-features` run below.
- **`9440d18` chore(deps)**: PayKit pin → `a569ac52` (head of pay-kit#259, reworked session API, idle-close root fix). Cascade includes the two items tracked in pay-kit#259's body — `with_blockhash_cache` pass-through and the fallible fail-the-challenge build — plus an end-to-end queued-touch-discard regression test mirroring the review harness (queued `Touch` dequeued after lease release must be discarded, `close_after` unchanged). The pin targets an unmerged branch head on purpose; it gets one final re-point to the squash commit when pay-kit#259 merges.
- **`2a6b14f` fix(cli)**: the remaining 4/5 finding, verified accurate against the live catalog. The Gemini entry's `default` pricing variant is literally described as "Fallback for unlisted Gemini models" — runtime `resolve_variant` accepts and prices any model through it, while the picker's family intersection hid everything not substring-matching a named variant. Discovery now skips the intersection when a `default` sentinel is declared, and drops non-chat models by capability (`supportedGenerationMethods`) instead. Providers without a `default` (e.g. Model Studio's named Qwen families) intersect exactly as before, so the audio/image/embedding noise stays hidden.

Verification at `2a6b14f`: `cargo test --workspace --all-features` — 26 targets, **1750 passed, 0 failed** (includes the live-network smoke test); `cargo fmt --all --check` and `cargo clippy -p pay --all-features` clean; `cargo check -p pay` (default features) clean.

rust/crates/core/src/server/payment.rs
```diff
@@ -383,6 +383,17 @@ async fn settle_axum_delegated_response(
...
+/// Resolved per-unit price in token base units — the session challenge's
+/// `amount` field (price per unit of service).
+pub(crate) fn price_unit_base_amount(price: &metering::ResolvedPrice, decimals: u8) -> u64 {
+    let per_unit = price
+        .dimensions
+        .first()
+        .map(|d| d.price_usd / d.scale.max(1) as f64)
+        .unwrap_or(0.01);
+    ((per_unit * 10f64.powi(i32::from(decimals))).round() as u64).max(1)
+}
```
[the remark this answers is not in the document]
lgalabru: Verified — accurate. `resolve_currency` does accept `SOL` (9 decimals), and the reworked challenge API is the first path that flows a real USD unit price into session base units (the old session challenge passed `u64::MAX`), so a SOL-currency session would advertise and settle $0.01 as 0.01 SOL.

Context: the 1-token==$1 assumption is systemic to the gateway's price model (`subscription.rs:90`, `openapi.rs:331`, the pre-existing metering ceiling in `gate.rs`, `session.cap_usdc`), and the operator `currencies` map is literally keyed `usd`. Deployed specs settle exclusively in USD stablecoins, so there is no live exposure.

Fixed in d208a87 by taking the restrict-to-pegged option: server startup now fails closed when a session currency is not a recognized USD-pegged stablecoin (symbol or known mint — unknown raw mints are rejected rather than assumed pegged), and `price_unit_base_amount` documents the invariant. Exchange-rate conversion for non-pegged assets stays out of scope; x402/charge paths are unchanged.

rust/crates/core/src/server/session.rs
```diff
@@ -607,12 +668,24 @@ impl SessionLifecycleRunloop {
...
+                        else {
+                            return true;
+                        };
+                        result
+                    }
+                    None => self.persist_touch(&channel_id, touched_at_ms).await,
+                }
+                .map_err(|error| error.to_string());
                 if let Err(error) = &result {
                     tracing::warn!(
```
EfeDurmaz16, line 710: persist_touch (just below, line 703) computes the idle-close deadline from the global close_delay only and never reads the channel's negotiated idle_timeout_seconds. PayKit re-arms close_after to now plus the negotiated timeout on every accepted action, but SessionMpp::process calls touch_channel right after and the store's monotonic guard lets the longer close_delay deadline win, so a client-selected shorter option is advertised, persisted, and never honored, holding deposits up to the ten minute default. Clamping the touch deadline per channel, for example min(close_delay, state.idle_timeout_seconds * 1000) using the state touch_channel_lifecycle already returns, restores the negotiated contract; the opposite direction is already safe.

rust/crates/cli/src/commands/payer_proxy.rs
```diff
@@ -587,6 +649,41 @@ async fn proxy(State(state): State<Arc<PayerState>>, req: Request) -> Response {
     }
 }
```
EfeDurmaz16, line 657: The reusable use credential is cached (line 624) before the retry that carries the open credential, so a transport error or 5xx on that paid call leaves a use credential cached for a channel the gateway may never have opened; the next request then gets a challenge-less session_failed 402 that the preserve branch keeps forever. Pre-PR code cached the idempotent open credential and dropped the cache on unrecognized 402s, so it recovered. Cache the use credential only after the paid retry returns non-402, or make unrecognized session_failed rejections terminal.

rust/crates/cli/src/commands/payer_proxy.rs
```diff
@@ -587,6 +649,41 @@ async fn proxy(State(state): State<Arc<PayerState>>, req: Request) -> Response {
...
+}
+
+fn retryable_cached_session_error(body: &[u8]) -> bool {
+    let text = cached_session_error_text(body);
+    text.contains("concurrent channel update")
+        || text.contains("retry the request")
+        || text.contains("session_capacity_reserved")
+}
+
+fn terminal_cached_session_error(body: &[u8]) -> bool {
```
EfeDurmaz16, line 683: terminal_cached_session_error misses at least five terminal gateway messages that arrive as challenge-less 402s: unknown session channel, session credential echoes a challenge this server did not issue, session channel predates proof binding, use authentication does not match the proof bound at open, and use is only valid for operator-signed sessions. With the default in-memory store and per-process challenge binding secret, a plain gateway restart triggers the first two, and the preserve branch then replays the dead credential on every request until the payer proxy itself is restarted. Adding these to the terminal list cannot open a replacement over a live channel and restores the pre-PR self-healing.

rust/crates/cli/src/commands/mod.rs
```diff
@@ -1176,7 +1190,8 @@ fn enforce_session_cap(
             "session payment cap requires a decoded SessionRequest".to_string(),
         ));
     };
-    let required_micro = amount_as_stablecoin_micro(&request.cap, &request.currency)?;
+    let required = session_deposit_amount(request).unwrap_or("0");
```
EfeDurmaz16: enforce_session_cap validates session_deposit_amount(request).unwrap_or("0") while pay_session_and_retry signs suggested.unwrap_or(1_000_000).max(minimum).max(1), so the enforced and signed amounts diverge: a challenge omitting both deposit fields passes a 0.10 cap yet locks 1 USDC, and a minimum above suggested enforces the smaller value but deposits the larger. Computing the deposit once with the signing expression and enforcing that value restores the pre-PR invariant where the cap bounded what was actually signed.

rust/crates/types/src/metering.rs
```diff
@@ -810,13 +819,14 @@ impl SignerConfig {
...
-    /// meters successful responses and signs their cumulative settlement.
-    Delegated,
+    Client,
+    /// The client binds the gateway operator as the channel's voucher signer;
+    /// the operator meters successful responses and signs their cumulative
+    /// settlement while the client authenticates with a reusable payer proof.
+    Operator,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, JsonSchema)]
```
EfeDurmaz16, line 832: SessionSpec has no deny_unknown_fields and no aliases for the removed keys, so an existing config with settlement_authority: delegated boots without error as voucher_signer: client, silently changing who signs settlement vouchers (verified with serde_yml against head). The sibling config structs in this file already use deny_unknown_fields; adding it here, or erroring explicitly on the removed legacy keys, keeps the upgrade fail-closed.

rust/crates/core/src/server/session.rs
```diff
@@ -752,12 +825,7 @@ impl SessionLifecycleRunloop {
...
             .next_settlement
             .map(|deadline| deadline.saturating_duration_since(Instant::now()));
-        match (close, settlement) {
-            (Some(close), Some(settlement)) => Some(close.min(settlement)),
-            (Some(close), None) => Some(close),
-            (None, Some(settlement)) => Some(settlement),
-            (None, None) => None,
-        }
+        next_lifecycle_wakeup(close, settlement)
     }
```
EfeDurmaz16, line 866: Pre-existing rather than introduced here, noting for a follow-up: when the settle in operator_close_channel fails after close_requested_at is set, no automatic path retries it, since close_due_channels filters on close_requested_at.is_none() (line 845) and the watermark and reconcile scans skip close-pending channels, so the retrying-after-delay log is wrong and only an authenticated wire close recovers the channel. Dropping the is_none clause from the close scan filter would let the existing Close-already-requested tolerance finish the settle.

rust/crates/core/src/server/session.rs
```diff
@@ -1295,26 +1388,18 @@ impl SessionMpp {
...
-            request.recent_slot = Some(hint.slot);
+    /// Build a [`PaymentChallenge`] for a new session.
+    ///
+    /// `amount` overrides the advertised per-unit price (base units) when the
+    /// gate resolved an endpoint-specific price; `None` keeps the configured
+    /// default.
+    pub fn challenge(&self, amount: Option<u64>) -> Result<PaymentChallenge> {
+        let mut request = self
+            .server
+            .build_challenge_request()
```
EfeDurmaz16, line 1435: Follow-up: during an RPC outage past the 45s blockhash cache window every session 402 fails as 500 challenge_generation_failed, which also blocks resume actions that never consume the hints and suppresses the endpoint's other schemes. The pay-kit migration note wanted fail-the-challenge for new channels only; issuing a hint-less session challenge when build_challenge_request fails keeps resume alive while new opens still fail loudly.

rust/crates/core/src/client/session.rs
```diff
@@ -111,25 +119,82 @@ impl SessionHandle {
...
+                voucher: None,
+            });
+            return build_header(&self.challenge, &action);
+        }
+        let action = match final_increment {
+            Some(increment) if increment > 0 => session
+                .close_action(Some(increment))
+                .await
+                .map_err(|e| Error::Mpp(format!("Failed to build close action: {e}")))?,
+            _ => SessionAction::Close(ClosePayload {
```
EfeDurmaz16, line 158: Follow-up: close_header(None) on a client-signed channel that never sent a voucher signs a cumulative-0 voucher the pinned process_close always rejects, since the equal-watermark replay check needs a byte match with a stored signature that does not exist, and the nonzero case passes only through deterministic re-signing of the identical message. Accepting a freshly signed voucher at the current watermark server-side, or guarding the zero case here, plus an end-to-end test for client-mode close_header(None), would close it.

rust/crates/core/tests/surfpool_tests.rs
```diff
@@ -501,38 +463,161 @@ async fn push_session_full_flow() {
...
+                        "context": { "slot": 34 },
+                        "value": [{
+                            "slot": 34,
+                            "confirmations": null,
+                            "err": null,
+                            "confirmationStatus": "finalized",
+                            "status": { "Ok": null }
+                        }]
+                    }),
+                    "getAccountInfo" => serde_json::json!({
```
EfeDurmaz16, line 574: Follow-up: the canned RPC answers getAccountInfo with the pre-built channel regardless of the requested address and confirms any signature, and this test is now the only integration coverage of the server-broadcast open path, so a wrong-account or wrong-signature regression would still pass. One-line hardening per arm: assert the requested address equals the channel id, and the queried signature equals the one echoed from sendTransaction.

rust/crates/cli/src/commands/server/inference/providers/catalog.rs
```diff
@@ -718,6 +744,68 @@ fn models_from_pricing_variants(endpoints: &[pay_core::skills::Endpoint]) -> Vec
...
+        return advertised_models;
+    }
+    let mut configured_models = models_from_pricing_variants(endpoints);
+    if configured_models.is_empty() {
+        configured_models.extend_from_slice(fallback_models);
+    }
+    if configured_models.is_empty() {
+        return advertised_models;
+    }
+    advertised_models
```
EfeDurmaz16, line 799: Follow-up, minor: if a provider has no default pricing variant and none of its configured families match the advertised list, this filter returns an empty Vec with no fallback, unlike the configured_models.is_empty() guard above. Unreachable in shipped config and the picker degrades to a no-models-reported row, so low stakes; falling back to configured_models would be safer.

EfeDurmaz16 (changes requested):
The heartbeat and cancel core is solid and the earlier Greptile P1s look addressed, but five verified issues block this round: the negotiated idle timeout is never honored by the gateway touch path, the payer proxy's new preserve-on-unknown-402 semantics wedge it permanently in reachable cases including a plain gateway restart, the cached use credential can outlive a failed open, the session cap check no longer bounds the signed deposit, and legacy session YAML silently downgrades to client voucher signing. Details inline with five follow-up notes. Small body nit: the PR text still references PayKit pin 3524a11b while Cargo pins dc3a36ee.

rust/crates/core/src/server/session.rs
lgalabru: Fixed in bbb2ff0. `persist_touch` now reads the channel's negotiated `idle_timeout_seconds` before computing the deadline and clamps to `min(close_delay_ms, idle_timeout_seconds * 1000)`. Red test: `touch_honors_negotiated_idle_timeout_shorter_than_close_delay` (session.rs), confirmed failing pre-fix. Also exposed `idle_timeout_options_seconds` in playground-api.yaml as a client-facing example now that it's actually honored.

rust/crates/cli/src/commands/payer_proxy.rs
lgalabru: Fixed in b86e3ca. The retry's credential is now only adopted into the cache once send_upstream proves non-402 — on a transport error the cache is left exactly as it was before the attempt, never adopting the unconfirmed credential. Updated the existing test that encoded the old (buggy) assumption: renamed session_open_credential_survives_a_paid_retry_transport_error to _does_not_survive_ and flipped its assertion to None, confirmed it fails against the old code.

rust/crates/cli/src/commands/payer_proxy.rs
lgalabru: Fixed in b86e3ca. Pinned the five terminal error texts as pub const in pay_core::server::session::terminal_errors (used at all five error-construction sites in session.rs) and added them to terminal_cached_session_error's checks, so the two sides can't drift apart again. Test grid added: terminal_cached_session_error_recognizes_every_server_rejection (one case per message) and a negative control (terminal_cached_session_error_does_not_misclassify_a_transient_store_error) for a plain store-read failure, which must stay retryable.

rust/crates/cli/src/commands/mod.rs
lgalabru: Fixed in 1004c4b. Extracted resolve_session_deposit as the single source of truth for the deposit amount (suggested, floored by minimum, falling back to 1 USDC) and used it in both enforce_session_cap and pay_session_and_retry. Red test: enforce_session_cap_bounds_the_exact_amount_that_gets_signed — a challenge with neither field set now correctly rejects a $0.10 cap instead of passing it.

rust/crates/types/src/metering.rs
lgalabru: Fixed in 6c9c1ea. Added deny_unknown_fields to SessionSpec. This also surfaced a live instance of exactly this bug: web-ui/proxy/alibaba-qwen.yml and google-gemini.yml were still on settlement_authority: delegated (pre-9440d18), silently booting as voucher_signer: client instead of operator. Migrated both to voucher_signer: operator in the same commit and verified all three shipped session configs (these two + playground-api.yaml) parse cleanly under the new attribute.

rust/crates/core/src/server/session.rs
lgalabru: Agreed, tracked as a follow-up rather than fixed in this round. Flagging the pattern match: this is the Rust twin of pay-kit#259's close re-drive blocker (close-pending channels invisible to every retry scan because of the same is_none()-style filter) — that fix's pattern (drop the filter / re-drive stranded settles) should port directly. Noted in the PR body.

rust/crates/core/src/server/session.rs
lgalabru: Legitimate, tracked as a follow-up, not addressed in this round. Noted in the PR body's review-response section.

rust/crates/core/src/client/session.rs
lgalabru: Legitimate, tracked as a follow-up, not addressed in this round. Noted in the PR body's review-response section.

rust/crates/core/tests/surfpool_tests.rs
lgalabru: Agreed — the canned RPC's blanket getAccountInfo/signature matching is a real coverage gap now that this is the only integration coverage of the server-broadcast open path. Tracked as a follow-up, not addressed in this round.

rust/crates/cli/src/commands/server/inference/providers/catalog.rs
lgalabru: Agreed, minor and unreachable in shipped config today, but worth the guard for defense-in-depth. Tracked as a follow-up, not addressed in this round.

EfeDurmaz16 (approved):
All five blockers are properly fixed at adc8e4d2, traced in code: persist_touch now clamps the deadline to the negotiated idle timeout, the retry credential is only adopted after a non-402 upstream response, the five terminal rejection texts are pinned as shared consts with a classifier test covering every server rejection, resolve_session_deposit is the single source for both the cap check and the signed amount, and SessionSpec rejects unknown fields (nice catch on the two live web-ui configs). Tracked follow-ups in the threads are fine as scoped, the close re-drive twin especially. CI is green.

PR #417, opened 2026-07-31
fix(proxy): default session close delay to ten minutes

lgalabru:
## Summary

- default omitted `session.close_delay_ms` values to 600,000 ms (ten minutes)
- update the public field documentation and lifecycle-default contract test

Explicit provider values continue to override this default.

## Validation

- `cargo fmt --check`
- `cargo test -p pay-types session_spec_defaults_lifecycle_intervals`

Related rollout: solana-foundation/agent-gateway#13


lgalabru:
Folded into #416 as commit 0a90373 so the session adjustments ship in one PR.

PR #420, opened 2026-08-02
build(deps): bump pay-kit to 16777d46

lgalabru:
Bumps pay-kit pin `dc3a36ee` → `16777d46` (one commit ahead: #265, `fix(rust/x402)!: fetch blockhash + openSlot via RPC when upto challenge omits hints`).

Breaking change: `build_upto_payload`/`build_upto_header` gained a required `&RpcClient` param (fallback when the 402 challenge omits `recentBlockhash`/`recentSlot` hints). Updated the one caller in `core/src/client/x402.rs` to pass the `RpcClient` already constructed for the upto path — same pattern the `exact` scheme in the same file already uses.

**Test plan**: `cargo test --workspace --all-features` → 1762 pass; `cargo clippy --workspace --all-targets --all-features` clean; `cargo fmt --check` clean.


EfeDurmaz16 (approved):
LGTM! 🚀

PR #422, opened 2026-08-03
feat: add pay-api and pay-worker as first-party crates

lgalabru:
## Summary
- Moves `pay-api` (Solana stablecoin balance/onramp HTTP service, confidential-transfer charges) and `pay-worker` (`close-channels` operator maintenance job, `settle-sessions` continuous session-settlement worker) in from `solana-foundation/agent-gateway`, where they lived as standalone Cargo workspaces pinning `pay-kit` independently.
- Both are now regular members of this workspace, sharing the single `pay-kit` pin already used everywhere else here (same rev they were already on — no wire-contract drift).
- `solana_mpp::` source references renamed to `pay_kit::` to match this repo's dependency alias.
- Adds `Dockerfile`s (build context is the whole `rust/` workspace, since both are workspace members now) and GHCR publish workflows, mirroring how the `pay` CLI image itself is published.

## Companion PR
solana-foundation/agent-gateway#17 removes the source from that repo and repoints its Docker-publish workflow to mirror these images from GHCR into its own Artifact Registry — no Terraform changes needed there, since the Artifact Registry image paths stay the same.

## Notes
- No behavior change — all 134 existing tests pass unmodified.
- Fixed the clippy/fmt lints this repo's stricter CI gate (`-D warnings`) newly surfaces on this code (agent-gateway's CI never ran clippy against it): collapsible-if, manual `is_multiple_of`, field-reassign-with-default, one genuinely-dead field (`#[allow(dead_code)]`'d with a note).
- Dropped the per-crate release profile tuning (thin LTO, symbol strip, `panic=abort`) agent-gateway had — profiles are workspace-level in Cargo, and applying it repo-wide would change release characteristics for the `pay` CLI binary too. Flagging as a deliberate follow-up if we want it back.
- Checked the moved code and its history for secrets before the move (private keys, API keys, hardcoded credentials) — found none; all sensitive values are Doppler-referenced by name, never committed.

## Test plan
- [x] `cargo test --workspace --exclude pay-integration` passes
- [x] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo fmt --check` clean
- [x] `docker build -f rust/crates/pay-api/Dockerfile rust` succeeds, binary runs (fails closed on missing config, as expected)
- [x] `docker build -f rust/crates/pay-worker/Dockerfile rust` succeeds, both binaries run (fail closed on missing config, as expected)


rust/crates/pay-api-core/src/receipt.rs
```diff
@@ -0,0 +1,3145 @@
...
+    account_keys: &[String],
+    program_id: Option<&str>,
+) -> Option<String> {
+    let program_id = program_id?;
+    for ix in top {
+        let program = ix.get("programId").and_then(Value::as_str).unwrap_or("");
+        if program != program_id {
+            continue;
+        }
+        let accounts = match ix.get("accounts").and_then(Value::as_array) {
```
lgalabru: This still fails both enforced clippy jobs (`clippy::question_mark`). Use `let accounts = ix.get("accounts").and_then(Value::as_array)?;` and collect from it; then rerun the exact workspace clippy command from CI.

lgalabru (commented):
CI is currently blocked on clippy.

rust/crates/pay-worker/src/bin/settle_sessions.rs
```diff
@@ -0,0 +1,1524 @@
...
+        )));
+    }
+    let signature = state
+        .highest_voucher_signature
+        .as_deref()
+        .ok_or_else(|| JobError::TxBuild("latest unsettled voucher has no signature".into()))?;
+    let expires_at = state
+        .highest_voucher_expires_at
+        .ok_or_else(|| JobError::TxBuild("latest unsettled voucher has no expiry".into()))?;
+    if expires_at <= now {
```
EfeDurmaz16: `expires_at == 0` is the protocol's never-expires sentinel, not an expiry in the past. pay-kit at the pinned rev 16777d4 documents it (`core/voucher.rs:44`) and writes `Some(0)` itself (`mpp/server/session.rs:1003`), so a session whose client omitted `expiresAt` never settles here. Once `close_requested_at` is set this returns `Err` every tick and the escrow stays locked, while the same settle would have succeeded on chain. Same comparison at line 952.

rust/crates/pay-worker/src/bin/close_channels.rs
```diff
@@ -0,0 +1,714 @@
...
+    let mint = decoded.mint();
+    let token_program = channel::resolve_token_program(rpc, rpc_url, &mint).await?;
+
+    let mut steps = Vec::new();
+
+    match decoded.channel.status {
+        STATUS_OPEN => {
+            if me == payee {
+                steps.push(PlannedStep {
+                    kind: StepKind::SettleAndSeal,
```
EfeDurmaz16: This seals at the on-chain settled watermark with `has_voucher = 0` and never reads the Redis session state, so any accepted-but-unsettled voucher value is forfeited back to the payer. `settle_sessions.rs:1099` applies the highest voucher first; the two jobs disagree and the destructive one has no guard. The documented `DRY_RUN=false just close-channels` runbook is what triggers it.

rust/crates/pay-api/src/signer.rs
```diff
@@ -0,0 +1,55 @@
...
+
+const LOCAL_PRIVATE_KEY_ENV: &str = "LOCAL_FEE_PAYER_PRIVATE_KEY";
+
+/// Build a fee-payer `SolanaSigner` from config + env. See module docs.
+pub async fn build_fee_payer_signer(
+    fee_payer: &FeePayerConfig,
+    missing_key_name_msg: &str,
+    missing_pubkey_msg: &str,
+) -> Result<Arc<dyn SolanaSigner>, Error> {
+    if let Ok(key) = std::env::var(LOCAL_PRIVATE_KEY_ENV) {
```
EfeDurmaz16, line 35: This takes precedence over the KMS signer unconditionally, with no network gate and no warning log, and the resulting pubkey is never checked against the `send.fee_payer.pubkey` that boot validation enforces. A stale export in the shell that runs a mainnet job moves signing off KMS silently. Same shape in `pay-worker/src/signer.rs:25`.

rust/crates/pay-api/src/endpoints/send.rs
```diff
@@ -0,0 +1,1470 @@
...
+        .request
+        .decode()
+        .map_err(|_| ApiError(Error::InvalidPaymentCredential))?;
+
+    validate_paid_send_request(&charge_request, &resolved).map_err(ApiError)?;
+
+    // Confidential settlement runs on the worker for the resolved request
+    // network (shared store for that network's orphan guard + replay
+    // protection); normal settlement stays on the direct per-request path.
+    let receipt = if resolved.confidential {
```
EfeDurmaz16, line 218: The settlement route is chosen from `SendRequest.confidential`, an unauthenticated body field, while `validate_paid_send_request` branches on the HMAC-signed `methodDetails.confidential`, and nothing cross-checks the two. A credential minted with `confidential: true` and submitted with `confidential: false` still passes validation and settles off the per-network worker that the comment above says owns the orphan guard and replay protection. Rejecting `resolved.confidential != (method_details.confidential == Some(true))` closes it.

rust/crates/pay-api/src/endpoints/subscriptions.rs
```diff
@@ -0,0 +1,896 @@
...
+
+    let signer = fee_payer_signer(&state)
+        .await
+        .map_err(|_| ApiError(Error::FeePayerSigner))?;
+    let mpp = new_mpp(&state, &resolved, Some(Arc::clone(&signer)))
+        .await
+        .map_err(ApiError)?;
+
+    // ── Verify the USDC charge credential ───────────────────────────────
+    let charge_receipt = match mpp.verify(&credential, &charge_request).await {
```
EfeDurmaz16, line 253: The charge settles here before anything checks the submitted tx is co-signable. `parse_cancel_tx` never calls `sanitize()` or `verify()` and does not check the signature count, so an empty `signatures` vec reaches line 571 and fails after the money moved: the caller gets a 5xx with the fee taken, nothing cancelled, and a credential that cannot be retried. Moving the signature-count and fee-payer-index checks into `parse_cancel_tx` puts them ahead of the charge.

rust/crates/pay-api-core/src/receipt.rs
```diff
@@ -0,0 +1,3147 @@
...
+    }
+    if subscription.period_amount.is_none() && state.amount_per_period > 0 {
+        subscription.period_amount = Some(ReceiptAmount {
+            asset: state.mint.to_string(),
+            symbol: None,
+            name: None,
+            logo_uri: None,
+            decimals: 0,
+            raw_amount: state.amount_per_period.to_string(),
+            ui_amount: state.amount_per_period as f64,
```
EfeDurmaz16, line 354: `ui_amount` is set to the raw base-unit value with `decimals: 0` above. The fixup in `attach_subscription_metadata` needs this mint present in the metadata map, but that map is built from `collect_mints`, which only walks the skeleton's transfers, total and account creations, and `state.mint` is not added until this line. A subscription receipt whose transaction carries no transfer in that mint reports 5000000.0 where it means 5.0.

.github/workflows/pay-api-docker.yml
```diff
@@ -0,0 +1,50 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - uses: docker/metadata-action@v5
+        id: meta
+        with:
+          images: ${{ env.IMAGE }}
+          tags: |
+            type=sha,prefix=sha-
+            type=raw,value=latest
```
EfeDurmaz16: `type=raw,value=latest` has no `enable={{is_default_branch}}` guard while the push condition below accepts `workflow_dispatch` with `inputs.push`, and `actions/checkout@v5` uses the dispatched ref. Anyone with write access can publish `:latest` from an unmerged branch. Same at `pay-worker-docker.yml:46`.

EfeDurmaz16 (commented):
Reviewed the moved crates against the money-path bar rather than as a pure relocation, since they become first-party here. Seven inline notes; the settle-sessions expiry sentinel and the close-channels seal are the two that fire on their own without an attacker. All pre-existing from agent-gateway, so a follow-up PR is the right home for them rather than this one.

Verified at 2cc1317 in a fresh clone: `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test --workspace --exclude pay-integration` 1900 passed 0 failed.

Redemption findings are covered separately and are not repeated here.

rust/crates/pay-api/src/endpoints/redeem.rs
```diff
@@ -0,0 +1,647 @@
...
+        cursor = txs.last().map(|t| t.signature.clone());
+        if page + 1 == cfg.max_scan_pages {
+            warn!(
+                code = code,
+                max_scan_pages = cfg.max_scan_pages,
+                "redeem scan hit page cap without exhausting hot-wallet history"
+            );
+        }
+    }
+    Ok(None)
```
EfeDurmaz16: Page-cap exhaustion returns the same value as a clean scan, so the handler at line 138 reads an unfinished scan as "never redeemed" and pays again. The scanned address is `send.fee_payer`, shared with `/v1/send`, so roughly 800 unrelated transactions after a burn is enough to re-open every spent code, and line 294 logs a warning and pays anyway. Suggest a distinct exhausted variant that fails closed with 502, matching the request-error branch at 139.

rust/crates/pay-api/src/endpoints/redeem.rs
```diff
@@ -0,0 +1,647 @@
...
+        &hot_wallet,
+        cfg,
+        blockhash,
+    ) {
+        Ok(tx) => tx,
+        Err(msg) => return err_resp(StatusCode::INTERNAL_SERVER_ERROR, &msg),
+    };
+
+    // 3. Sign + broadcast.
+    let signature = match sign_and_broadcast(unsigned, signer, &state, &rpc_url).await {
```
EfeDurmaz16: The burn is never recorded before broadcast, and `InFlightGuard` drops when the handler returns, so the only surviving dedup is the Helius scan, which cannot see an unconfirmed or unindexed transaction. A second POST with the same code a second later gets a fresh blockhash, a different signature, and a second payout; the route has no auth or rate limit. Commit 1ee954b's message claims durability across instances and restarts, but a process-local `Mutex` plus an indexer read-back gives neither. A durable record keyed by code, written before broadcast, is what the message describes.

EfeDurmaz16 (commented):
The relocation itself is clean: `rust/Cargo.toml` needed no edit because `members = ["crates/*"]` already covers the new directories, `Cargo.lock` is additive only so no existing pin moved, and `default-members = ["crates/cli"]` keeps the new `gcp_kms`/`worker`/`redis-store` features out of the released CLI binary. Three inline notes below, all on redemption, all pre-existing in agent-gateway rather than introduced here. Sting covers settlement, send, subscriptions and CI separately.

The theme: commit 1ee954b makes the API key mandatory so operators believe dedup is on, but the two paths that silently return "never redeemed" are still open, so the failure got quieter rather than safer. Worth splitting into a follow-up PR so this one stays a pure move.

Verified at 2cc1317 in a fresh clone: read `redeem.rs`, `config.rs`, `main.rs`, `state.rs` in full, diffed `Cargo.lock` and `Cargo.toml` against merge-base `e1172d4`, and traced each finding to the payout call. I could not diff the moved sources against agent-gateway, so "no behavior change" is unverified from outside.

rust/crates/pay-api/config/default.yaml
```diff
@@ -0,0 +1,85 @@
...
+    memo: fee-payer-refund
+
+# Activation-campaign redemption. Reuses `send.fee_payer.*` as the hot
+# wallet that signs Transfer + Memo. A Helius API key is required for
+# durable redemption deduplication across API instances and restarts.
+redemption:
+  enabled: true
+  currency: USDC
+  amount: 5000000          # $5 legacy flat-list fallback; campaigns override it
+  network: sandbox         # → networks.sandbox.rpc_url
```
EfeDurmaz16: The shipped default enables redemption on sandbox, but the dedup scan always hits `helius_base` (`config.rs:304`, `https://api.helius.xyz`) with no network parameter, so a sandbox burn can never match and `find_prior_burn` returns `None` on every request. This is the same hole the removed sandbox escape hatch used to announce with a warning. Either derive the scan endpoint from `redemption.network` or reject the mismatch at boot.

rust/crates/pay-api/src/signer.rs
lgalabru: Fixed in 0ca0114 in both pay-api and pay-worker. A local signer is accepted only when its derived pubkey exactly matches the configured fee-payer pubkey, and selecting the local-key path now emits a warning. A stale or mismatched export fails closed.

rust/crates/pay-api/config/default.yaml
lgalabru: Fixed in 0ca0114. Redemption validation now rejects every non-mainnet network because the configured Helius history endpoint is mainnet-only. The shipped default is disabled and uses mainnet, so sandbox cannot silently bypass deduplication.

rust/crates/pay-api/src/endpoints/subscriptions.rs
lgalabru: Fixed in 0ca0114. `parse_cancel_tx` now sanitizes the transaction and validates the required signature-slot count before charge verification. Added a regression test for an empty signatures vector.

rust/crates/pay-worker/src/bin/settle_sessions.rs
lgalabru: Fixed in 0ca0114. Both voucher-expiry checks now treat `expires_at == 0` as the protocol's never-expires sentinel, and the idle-close test covers a zero-expiry voucher at `i64::MAX`.

rust/crates/pay-api/src/endpoints/redeem.rs
lgalabru: Fixed in 0ca0114. The process-local guard was replaced with a required Redis claim store. Redemption performs an atomic `SET ... NX` before building/broadcasting, never deletes the claim, records the broadcast signature afterward, and fails closed when Redis is unavailable. Claim keys hash the code rather than exposing it.

.github/workflows/pay-api-docker.yml
lgalabru: Fixed in 0ca0114 in both image workflows. The raw `latest` tag now uses `enable={{is_default_branch}}`; manual runs on non-default refs can publish only immutable SHA tags.

rust/crates/pay-api-core/src/receipt.rs
lgalabru: Fixed in 0ca0114. When recurring-delegation state introduces a mint that was absent from the receipt skeleton, the async builder resolves and merges that mint's metadata before formatting `period_amount`. Added a regression test verifying 5,000,000 base units renders as 5.0 with 6 decimals.

rust/crates/pay-worker/src/bin/close_channels.rs
lgalabru: Fixed in 0ca0114. `close-channels` now refuses to seal an open channel when running as payee because it has no Redis voucher state; it directs operators to `settle-sessions`, which applies the highest accepted voucher before sealing.

rust/crates/pay-api/src/endpoints/redeem.rs
lgalabru: Fixed in 0ca0114. Page-cap exhaustion is now a distinct `DedupScanError::PageCap` and the handler fails closed with 502 instead of treating an incomplete scan as no prior redemption. Zero-page configuration is also rejected at startup.

rust/crates/pay-api/src/endpoints/send.rs
lgalabru: Fixed in 0ca0114. Paid-send validation now requires the signed `methodDetails.confidential` value to match the resolved request mode before selecting the settlement path. Added a regression test for the true/false mismatch.

rust/crates/pay-worker/src/channel.rs
```diff
@@ -0,0 +1,479 @@
...
+    ))
+}
+
+/// Decode `count(u32 LE) || entries(count × 34)` into `DistributionEntry`s.
+fn decode_recipients(preimage: &[u8]) -> Result<Vec<DistributionEntry>, JobError> {
+    if preimage.len() < 4 {
+        return Err(JobError::OpenIxDecode("preimage shorter than count".into()));
+    }
+    let count = u32::from_le_bytes([preimage[0], preimage[1], preimage[2], preimage[3]]) as usize;
+    let mut out = Vec::with_capacity(count);
```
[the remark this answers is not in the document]
lgalabru: Fixed in dbb379f.  now rejects counts above the protocol-safe 64-recipient limit and validates the complete payload length before calling . It also rejects trailing bytes. Added a regression test for a four-byte  count; pay-worker focused tests and clippy pass.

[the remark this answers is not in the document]
lgalabru: The fix is in dbb379f. The parser now bounds the count to 64, checks the exact payload length before allocation, rejects trailing bytes, and includes a regression test for a u32::MAX count. Focused pay-worker tests and clippy pass.

rust/crates/pay-api/src/endpoints/redeem.rs
```diff
@@ -0,0 +1,671 @@
...
+
+    // 3. Build the unsigned tx.
+    let rpc_url = match state.rpc_url_for(cfg.network) {
+        Ok(url) => url.to_string(),
+        Err(e) => return err_resp(StatusCode::INTERNAL_SERVER_ERROR, &e.to_string()),
+    };
+    let blockhash_str = match state.rpc.get_latest_blockhash(&rpc_url).await {
+        Ok(b) => b,
+        Err(e) => return err_resp(StatusCode::BAD_GATEWAY, &e.to_string()),
+    };
```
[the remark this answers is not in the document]
lgalabru: Fixed in 069b77f. Claims remain retryable through blockhash lookup, transaction construction, signing, and serialization failures; those paths atomically release the claimed:<destination> value. Immediately before sending, the claim transitions atomically to broadcasting:<destination> and is retained for ambiguous RPC failures. Successful sends record the signature. Focused pay-api tests and clippy pass.

rust/crates/pay-worker/src/bin/close_channels.rs
```diff
@@ -0,0 +1,714 @@
...
+                // hard failure when we were actually trying to broadcast.
+                if dry_run {
+                    warn!(channel = %address, error = %err, "channel error (dry-run)");
+                } else {
+                    error!(channel = %address, error = %err, "channel error");
+                    hard_failures += 1;
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
lgalabru: Address-scoped Redis leases are now acquired before each channel is planned. The lease is held through reclaim batching and released atomically by owner after completion; a 300-second TTL covers worker crashes. Broadcasting mode now requires PAY_SESSION_REDIS_URL.

.github/workflows/pay-api-docker.yml
```diff
@@ -0,0 +1,50 @@
...
+      - uses: docker/build-push-action@v6
```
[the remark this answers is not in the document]
lgalabru: Fixed in both pay-api-docker.yml and pay-worker-docker.yml. checkout, login, metadata, and build-push are pinned to reviewed full commit SHAs, with the release tags retained as comments for traceability.

rust/crates/pay-api/src/endpoints/redeem.rs
```diff
@@ -0,0 +1,759 @@
...
+        }
+    }
+    let signature = match broadcast_transaction(&state, &rpc_url, &signed_tx).await {
+        Ok(sig) => sig,
+        Err(msg) => {
+            // The claim is intentionally retained: the RPC may have accepted
+            // the transaction even when its response was lost.
+            return err_resp(StatusCode::BAD_GATEWAY, &msg);
+        }
+    };
```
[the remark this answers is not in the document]
lgalabru: Fixed. The broadcast path now preserves the RPC error type: recognized deterministic simulation/preflight failures atomically release the claim for retry, while transport, timeout, malformed-response, and otherwise ambiguous failures retain the broadcasting claim.

EfeDurmaz16 (commented):
Reviewed the follow-up commits rather than the move. dd21394 is production code from agent-gateway, and the relocation itself checks out: the shipped CLI feature graph is identical to main, Cargo.lock is additive with no pin moved, and no newly enabled feature changes behavior.

The four items below are different. They were introduced by this PR's own fix commits, not inherited from agent-gateway, so they are not covered by the "this is a move" framing.

Everything else I found is pre-existing in agent-gateway and I would rather not hold a relocation on it. Happy to open a follow-up PR with those and approve this one once the four below are settled. One of them, an uncapped priority fee on the operator-signed cancel transaction, is cheap to exploit, so I would not want it to land quietly in a backlog.

Separately, worth adding a redis service to the Rust CI jobs before merge. The claim store and both leases added during this review have never executed in a test, so the green run does not cover them.

rust/crates/pay-api/src/endpoints/redeem.rs
```diff
@@ -0,0 +1,778 @@
...
+            return err_resp(StatusCode::BAD_GATEWAY, "redemption claim failed");
+        }
+    }
+    let signature = match broadcast_transaction(&state, &rpc_url, &signed_tx).await {
+        Ok(sig) => sig,
+        Err(error) if is_deterministic_preflight_error(&error) => {
+            // A rejected preflight never submits the transaction, so the
+            // claim can be retried. Transport and timeout errors remain
+            // claimed because the RPC may have accepted the transaction.
+            release_redemption_claim_best_effort(cfg, &req.code, &destination_pk).await;
```
EfeDurmaz16, line 212: This release cannot fire. `mark_redemption_broadcasting` at line 198 has already rewritten the value to `broadcasting:{destination}`, but the release script compares against `claimed:{destination}`, so the Lua returns 0 and the DEL never runs. `release_redemption_claim_best_effort` only logs on `Err`, so the `Ok(false)` is silent. With no TTL on the claim key the code is then locked permanently for a failure that provably sent no funds. Matching either prefix in the release script closes it.

rust/crates/pay-api/src/endpoints/redeem.rs
```diff
@@ -0,0 +1,778 @@
...
+                    return Ok(Some(tx.signature.clone()));
+                }
+            }
+        }
+        // Underfull page → end of history.
+        if txs.len() < HELIUS_PAGE_LIMIT {
+            return Ok(None);
+        }
+        cursor = txs.last().map(|t| t.signature.clone());
+        if page + 1 == cfg.max_scan_pages {
```
EfeDurmaz16, line 354: For a code with no prior burn the scan can only reach `Ok(None)` from an empty or underfull page, so proving absence means walking the hot wallet's history to the end. That wallet is `send.fee_payer`, shared with `/v1/send`, and every successful redemption appends to it, so once it passes `max_scan_pages * 100` transactions every fresh redemption returns 502. The fail-closed direction is right; bounding the scan at the claim-store cutoff keeps it without the cliff, since codes issued after that cutoff are already deduped by Redis.

web-ui/proxy/helius.yml
```diff
@@ -10,7 +10,7 @@ routing:
   auth:
     method: query_param
     key: "api-key"
-    value_from_env: HELIUS_API_KEY
+    value_from_env: SOLANA_RPC_API_KEY
```
EfeDurmaz16: This file predates the move by two months and its siblings all use provider-specific names (`DASHSCOPE_API_KEY`, `GOOGLE_API_KEY`), so this rename makes helius the only one off the pattern, for a proxy config that is specifically a Helius endpoint. It also means pay-api and `pay gate api` now read the same variable, so a host running both can no longer give them separate keys. If the rename is wanted anyway, worth doing the siblings in the same pass rather than one file.

rust/crates/pay-worker/src/bin/close_channels.rs
```diff
@@ -0,0 +1,805 @@
...
+        reclaim_candidates,
+        dry_run,
+        confirm_timeout,
+    )
+    .await;
+    acted += reclaim_outcome.acted;
+    hard_failures += reclaim_outcome.failures;
+    for lease in reclaim_leases {
+        lease.release().await;
+    }
```
EfeDurmaz16, line 348: Reopening greptile's thread here since it was resolved without a fix commit or a reply. The lease is acquired with a fixed 300s TTL, held across the whole channel loop and then the reclaim batches, and never renewed, while each confirmed broadcast can take up to `confirm_timeout` (60s default), so a handful of channels exceeds it. The same shape applies to the settlement lock and the redemption claim, which are also acquire-once with no renewal. If this is a deliberate accept, worth saying so on the thread.

rust/crates/pay-api/src/endpoints/redeem.rs
lgalabru: Fixed in 22223a9. The owner-checked release script now deletes either the pre-broadcast `claimed:<destination>` value or the post-signing `broadcasting:<destination>` value, so a deterministic preflight rejection can be retried.

rust/crates/pay-api/src/endpoints/redeem.rs
lgalabru: Fixed in 22223a9. Redemption now requires `legacy_scan_cutoff_unix_seconds`, the Redis claim-store rollout timestamp. The Helius scan stops once it reaches that boundary, retaining legacy memo detection beforehand while relying on durable Redis claims afterward instead of failing every fresh redemption once the hot wallet grows.

web-ui/proxy/helius.yml
lgalabru: Reverted in 22223a9. The Helius proxy is back on `HELIUS_API_KEY`, keeping its provider-specific credential and preserving separate keys for the proxy and pay-api.

rust/crates/pay-worker/src/bin/close_channels.rs
lgalabru: Fixed in 22223a9. Reclaim leases now use an owner-checked `EXPIRE` heartbeat every third of the TTL and cancel/await it before owner-checked release. I applied the same renewal to the settle-sessions lock. The redemption claim itself intentionally has no TTL, so it does not have this expiry race.

EfeDurmaz16 (approved):
@lgalabru 

Approved, previous blockers look fixed at 22223a9: claim release, scan cutoff, Helius env revert, and lease heartbeats. 

Please track follow-ups before enabling cancel, redeem, or forced-close in prod: cap cancel priority fees, settle vouchers during CLOSING, and add Redis to Rust CI. Fine to merge the move with those tracked and all commits are signed.

PR #426, opened 2026-08-10
Fix server scaffold to use current routing schema

epistemedeus:
## Summary

- update `pay server scaffold` to emit the current `routing` proxy schema
- add a regression test that parses the generated template as `ApiSpec`
- verify the parsed route is `RoutingConfig::Proxy`

## Reproduction

With pay CLI 0.27.0:

```sh
pay server scaffold /tmp/paywall.yml
pay server start /tmp/paywall.yml --no-register
```

The generated file uses the retired top-level `forward_url` field, so the server cannot parse its own scaffold. This change emits:

```yaml
routing:
  type: proxy
  url: https://api.example.com
```

## Verification

- `cargo fmt --check`
- `cargo test -p pay-cli scaffold_template_uses_the_current_routing_schema`

The targeted test passes.

### DIFF
--- a/rust/crates/cli/src/commands/server/scaffold.rs
+++ b/rust/crates/cli/src/commands/server/scaffold.rs
@@ -14,7 +14,9 @@ title: "My API"
 description: "API description"
 category: ai_ml
 version: v1
-forward_url: https://api.example.com
+routing:
+  type: proxy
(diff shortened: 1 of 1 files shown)


lgalabru (approved):
thanks @epistemedeus!

PR #435, opened 2026-08-14
feat(bench): consolidate push and benchmark slices

lgalabru:
## Summary
- consolidate the Pay push pipeline, sponsored session opens from #430, and high-throughput Pingora benchmark work
- add bounded-concurrency CSV distributions, durable journals, deterministic MPP-session fixtures, and worker-owned load sources
- deploy devnet-only Token-2022 USDtest with public metadata and retain the 100k-wallet derivation set
- permit an explicitly configured, opt-in Token-2022 benchmark mint for mainnet session simulation; validate its mint owner at gateway startup and propagate the program to session channels
- pin pay-kit PR #288 on top of #283 and add reproducible Ubuntu/Ansible deployment and host tuning

## Test Plan
- cargo fmt --all -- --check
- cargo test -p pay session_currencies_require_usd_pegged_stablecoins
- prior PR suite: cargo clippy -p pay -p pay-core -p pay-bench --all-targets -- -D warnings
- prior PR suite: cargo test -p pay-core -p pay -p pay-bench

## Live devnet verification
- bounded fanout: 98,720 recipients, 10,969 confirmed transactions, 0 failed/unknown
- sponsored 100-user session smoke: 2,646/2,646 dispatched requests returned 200; 0 failures; all channels cooperatively closed


rust/bench/configs/session-devnet.yml
```diff
@@ -0,0 +1,32 @@
...
+
+load:
+  users: 100
+  requests_per_sec_per_user: 5
+  prepare_secs: 5
+  unleash_secs: 15
+  max_concurrency: 512
+
+endpoints:
+  - url: http://213.239.141.29:1402/api/v1/compute
```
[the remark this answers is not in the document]
lgalabru: Valid. `validate_payment_transport` rejects this public HTTP URL before provisioning. We will move the deployed fixture to HTTPS; the transport guard should stay strict.

rust/bench/src/config.rs
```diff
@@ -0,0 +1,355 @@
...
+        if self.load.shard_count == 0 || self.load.shard_index >= self.load.shard_count {
+            bail!("load.shard_index must be less than non-zero load.shard_count");
+        }
+        if self.endpoints.is_empty() {
+            bail!("at least one endpoint is required");
+        }
+        self.tls_ca_certificate()?;
+        if self.run.safety.max_total_usdc < 0.0 || self.run.safety.max_total_sol < 0.0 {
+            bail!("safety caps must be >= 0");
+        }
```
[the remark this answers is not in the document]
lgalabru: Fixed in b4b61a75. Confirmed real — and `SetupConfig` (fixtures.rs) already guards its cap with `is_finite() && >= 0.0`; `run.safety` was the outlier.

Two-layer fix:
- **Validation** (`config.rs`): each cap must now be `is_finite() && >= 0.0`, so `.nan`/`.inf`/`-.inf` bail at load.
- **Enforcement** (`engine.rs`): `enforce_caps` now compares via `partial_cmp` (`None => reject`) instead of `total > cap`, so a non-finite cap *or* estimate fails closed even if validation were bypassed.

Added coverage for `.nan`/`.inf`/`-.inf` rejection and the enforce-caps fail-closed path (33/33 bench tests green, clippy clean).

rust/bench/deploy/pay-bench-devnet.service
```diff
@@ -0,0 +1,19 @@
...
+Description=Pay Bench devnet MPP-session Pingora gateway
+Wants=network-online.target
+After=network-online.target
+
+[Service]
+Type=simple
+User=ubuntu
+WorkingDirectory=/home/ubuntu/pay-bench/rust
+EnvironmentFile=/etc/pay-bench/devnet.env
+ExecStart=/home/ubuntu/pay-bench/rust/target/release/pay server start /home/ubuntu/pay-bench/rust/bench/configs/devnet-pingora.yml --bind 0.0.0.0:1402 --no-register
```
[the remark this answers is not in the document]
lgalabru: Valid. The live host is protected by a native-TLS systemd override, but a fresh install from this unit would be unsafe. I changed the unit to require the certificate and key and documented their paths; it will ship with the pending signed stack.

rust/crates/core/src/server/session.rs
```diff
@@ -1708,6 +2015,95 @@ impl SessionMpp {
...
+        let Some(chan) = self
+            .operator_runtime
+            .fetch_payment_channel(channel_id)
+            .await?
+        else {
+            return Ok(());
+        };
+        // Only adopt channels still open (status 0); sealed/closing ones cannot
+        // accept new vouchers.
+        if chan.status != 0 {
```
EfeDurmaz16: Only gate is `status == 0`, so a client can point a voucher at a channel whose payee and mint it controls. Check both against config before the insert?

rust/crates/cli/src/commands/server/start.rs
```diff
@@ -769,18 +812,21 @@ impl StartCommand {
...
                 Some(signer)
             } else if let Some(ref cfg) = signer_cfg {
                 Some(resolve_signer(cfg).await?)
-            } else if let Some(signer) = super::load_account_or_legacy_signer(
-                network.slug(),
-                account_override.as_deref(),
-                legacy_signer_source.as_deref(),
-                &pay_core::keystore::AuthIntent::use_gateway_fee_payer(),
-            )? {
+            } else if (fee_payer || !has_explicit_recipient)
```
EfeDurmaz16: This also gates the settlement signer at 1109, so a session gateway with `operator.recipient` and `fee_payer` off takes vouchers and never settles, silently.

rust/crates/pay-api/src/endpoints/transfer_batches.rs
```diff
@@ -0,0 +1,132 @@
...
+    let runtime = TransferBatchRuntime::resolve(
+        &chunk,
+        &state.rpc,
+        &push.networks,
+        &push.sponsor,
+        &push.settings,
+    )
+    .map_err(ApiError)?;
+
+    match headers.get(AUTHORIZATION) {
```
EfeDurmaz16, line 53: No caller auth, and nothing collects `feeReimbursementRaw`, so the sponsor pays fees and ATA rent for anyone. Land billing before enabling this?

rust/crates/core/src/client/push/executor.rs
```diff
@@ -0,0 +1,1171 @@
...
+            .header(
+                "Authorization",
+                format!("Bearer {}", signed.signed_transaction_base64),
+            )
+            .json(&serde_json::json!({
+                "batchId": self.batch_id,
+                "chunkIndex": signed.chunk_index,
+                "sender": self.sender.to_string(),
+                "currency": self.currency,
+                "network": self.network,
```
EfeDurmaz16: `transfers` is `rowId` only but `TransferBatchEntry` needs `recipient` and `amount`, so this cannot deserialize. `prepare` also hardcodes 6 decimals.

rust/crates/core/src/client/push/journal.rs
```diff
@@ -0,0 +1,1016 @@
...
+            file,
+            path,
+            next_sequence: 1,
+        })
+    }
+
+    /// Re-open an existing journal for a resumed run. `next_sequence` picks
+    /// up after the highest sequence number [`load_events`] can recover
+    /// (tolerating one truncated final line).
+    pub fn open_existing(path: PathBuf) -> Result<(Self, Vec<JournalEvent>)> {
```
EfeDurmaz16, line 196: `load_events` skips a truncated last line without truncating the file, so this append merges into it. Truncate to the last good offset first?

rust/crates/cli/src/commands/fanout.rs
```diff
@@ -0,0 +1,606 @@
...
+    /// RFC 4180 CSV with exactly `recipient,amount` headers.
+    #[arg(value_name = "CSV")]
+    csv: PathBuf,
+
+    /// Stablecoin symbol (for example USDC, USDG, or devnet-only USDtest).
+    #[arg(long, value_name = "STABLECOIN")]
+    currency: String,
+
+    /// Solana network whose mint and account registry entry will be used.
+    #[arg(long, default_value = "mainnet")]
```
EfeDurmaz16, line 48: `mainnet` default, and the global `--sandbox` / `--mainnet` are ignored. `validate_network` keys off the slug, so `--network devnet` with a mainnet `--rpc-url` passes the keypair ban.

EfeDurmaz16 (commented):
Nice work. Three worth a look before merge, all inline: `reuse_from_chain` adopting a channel with no payee or mint check, the `start.rs` signer gate also dropping the settlement signer, and `transfer-batches` having no caller auth. Three smaller ones inline too.

EfeDurmaz16 (commented):
Thanks, the fixes look good. I'll approve once the two live Greptile P1s are closed (deferred settlement failing for `close_after_run`: false, and the missing BENCH_TLS_CA_CERT setup in the bench README) and e8335f0, 448b5ef, b1bb4c6 are signed. Those three carry Signed-off-by but no gpgsig, so GitHub shows them unverified. I know signing them means rebasing from that point, happy to wait if you'd rather batch it.

EfeDurmaz16 (approved):

PR #451, opened 2026-09-01
fix(mcp): align provider guidance with catalog format

JasonColapietro:
## Summary
- make create_skill accept the current provider directory layout ending in PAY.md
- emit current pay catalog check commands instead of removed pay skills commands
- update the bundled monetization reference for committed OpenAPI sidecars, current categories, and catalog validation
- remove the obsolete provider sync recommendation from the MCP tool description

## Problem
The MCP helper still generated and validated the legacy flat provider layout. A contributor following it would produce paths and commands rejected by the current pay-skills catalog workflow.

## Verification
- cargo test -p pay-mcp: 107 passed, 3 ignored
- cargo test -p pay-mcp --doc
- cargo fmt --all --check
- git diff --check

The output-path tests were updated first and failed against the old implementation, then passed with this change.

### DIFF
--- a/rust/crates/mcp/src/server.rs
+++ b/rust/crates/mcp/src/server.rs
@@ -157,14 +157,15 @@ a purchase; it only renders the QR PNG and returns the funding address.
 
 Use this when a developer wants to publish a payment-gated API in
 https://github.com/solana-foundation/pay-skills. Pass the complete provider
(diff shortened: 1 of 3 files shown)


lgalabru (approved):
LGTM. thanks @JasonColapietro!

PR #454, opened 2026-09-01
feat(x402): make batch-settlement a first-class scheme

lgalabru:
## Summary

Makes x402 `batch-settlement` a scheme `pay` can actually serve and pay, and picks up the pay-kit rewrite that matches the current [SVM specification](https://github.com/x402-foundation/x402/blob/main/specs/schemes/batch-settlement/scheme_batch_settlement_svm.md).

`Scheme::X402BatchSettlement` already existed in the enum and the gate already dispatched on it — but nothing implemented `PaymentState::x402_batch`, so no `pay` server could offer it, and the client had no way to pay one.

## Why it is shaped like a session, not like `upto`

Batch-settlement is stateful across requests: one escrow channel backs many calls and both sides track a cumulative watermark. That is the MPP session model, not the `upto` model, which opens and closes a channel per request.

| | mpp-session | x402-upto | batch-settlement |
|---|---|---|---|
| channel lifetime | many requests | one request | **many requests** |
| server state | `ChannelStore` (Redis) | none | **`ChannelStore` (Redis)** |
| client state | MCP `SessionCache` | none | **MCP `BatchChannelCache`** |

## Server

- `pay serve` builds the backend whenever an endpoint accepts the scheme and an operator signer is configured, sharing the session channel store. `PAY_X402_REDIS_URL` falls back to `PAY_MPP_REDIS_URL` and the legacy `PAY_SESSION_REDIS_URL`, under its own key prefix — both schemes store `ChannelState` keyed by channel id, so a shared namespace would let one read the other's channels. An in-memory store warns at boot: it holds the only record of what a client has been charged, and a restart forfeits every unclaimed voucher.
- The gate follows the scheme's `authorization` flow. Verification is read-only and reserves the channel; the voucher commits only after the upstream has served, so a failed request leaves the client uncharged and free to retry the same voucher. `GateDecision::Forward` carries the reservation, and both adapters — the axum middleware and the Pingora proxy — settle it in the post-response hook they already run for `upto`.
- A verification failure re-challenges with the server's corrective 402, which carries its watermark plus the client's own signed voucher proving it.
- A refund bypasses the upstream: a channel close is a payment-control operation, not a paid request.

## Client

- `BatchChannelCache` holds open channels for the life of the process. The MCP server keeps one per connection, beside its MPP session cache, so a long-lived connection reuses the same channel. It currently tops up one request at a time, and a larger deposit option can follow separately. A one-shot `pay curl` escrows exactly one request's price, recoverable by force-close.
- The watermark advances only when `PAYMENT-RESPONSE` confirms the exact commitment that was sent. Anything else leaves it alone, so the next request re-signs the same amount and the server sees the idempotent retry it is. On a corrective 402 the MCP path resynchronizes and retries once.

## A bug this found

A batch-only 402 carries the same core fields as `exact` (scheme, network, amount, asset, `payTo`), so `pay`'s lenient `exact` reader accepted it — the client would have tried to pay with an exact transfer that the server rejects. Classification now prefers the batch reading and drops the bogus `exact` alternative when a batch offer is on the same 402, the same guard `upto` already had. Pinned by a regression test.

## Refactor

Adding the second channel scheme duplicated the first, so three pairs are now one implementation each. No behavior change.

- `prepare_channel_payment` — the preamble both payment builders need (CAIP-2 to cluster, Surfpool detection, network-intent check, wallet load, RPC, auto-fund). It was 43 identical lines across the two; `build_upto_payment_with_override` drops 101 → 55 lines and `build_batch_payment` 147 → 107.
- `pay_channel_and_retry` — the CLI notice/build/retry/receipt path, with the scheme label and header builder passed in.
- `channel_store` — either scheme's Redis-or-memory resolver, taking the env vars, key prefix and label. The prefix was a convention to remember; it is now a parameter.

## Testing

1328 workspace tests pass; `clippy --workspace --all-targets` clean.

Not covered by unit tests: the RPC-dependent server paths (deposit broadcast, claim/settle/finalize/reclaim) and a live end-to-end channel. Those want a surfpool integration test — happy to add one here if you would rather not land it without.

## Known follow-ups

- `cli/src/commands/server/inference/mod.rs` has its own `PaymentState` impl and does not offer the scheme.
- Replaying a completed voucher (client lost the original response) returns settlement JSON, not the purchased resource — tracked in #455.

## Performance (SUNBURST)

Benchmarked at load as an A/B against mpp-session, and taken from **~42k → ~888k req/s** (2000-channel closed loop, AVX-512 host, p50 191ms → 2.0ms) — now **ed25519-verify-bound at mpp parity**. Three changes here plus the pay-kit clone-elimination land the win:

| change | where | throughput |
|---|---|---|
| glibc baseline | — | 42–57k/s |
| `mimalloc` global allocator (`ff4c6d72`) | this PR | 110k/s (2.3x) |
| eliminate per-request `ChannelState` clone | pay-kit solana-foundation/pay-kit#304 (bumped by `bcd0369b`) | **888k/s (8x)** |

The per-request deep clone of `ChannelState` was hammering the glibc arena locks (~33% of CPU in futex under load); `mimalloc`'s thread-local heaps absorbed the pressure, then removing the clone in pay-kit took it the rest of the way. See pay-kit#304 for the root-cause writeup.

Also in this PR for the benchmark:
- `rust/bench/src/scheme/batch_settlement.rs` — a `Scheme::X402BatchSettlement` load driver (pure-sync ed25519 voucher signing, `build_deposit` provisioning, `PAYMENT-SIGNATURE` header), wired through `bench/src/config.rs` + `scheme/mod.rs`.
- `PAY_BATCH_SNAPSHOT_MAX_AGE_SECS` env on `server start` to control snapshot staleness independent of the reconcile cadence.

This now depends on both solana-foundation/pay-kit#301 (the rewrite) and solana-foundation/pay-kit#304 (the perf fix), pinned by rev (`bcd0369b`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rust/crates/core/src/client/x402.rs
```diff
@@ -403,6 +476,167 @@ pub fn build_upto_payment_with_override(
...
+        ephemeral_notice,
+        rpc,
+        rt,
+    } = prepare_channel_payment(
+        ChannelOffer {
+            network: &requirements.network,
+            amount: &requirements.amount,
+            asset: &requirements.asset,
+            recent_blockhash: requirements.extra.recent_blockhash.as_deref(),
+        },
```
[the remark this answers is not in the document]
lgalabru: Addressed in 4d2e2384. The client now determines whether the request needs an open/top-up before it asks the wallet for authorization; that prompt uses the exact escrow amount and labels it channel escrow. A voucher covered by existing escrow continues to authorize only the request price.

rust/crates/cli/src/commands/server/start.rs
```diff
@@ -65,31 +90,62 @@ async fn session_channel_store()
...
-                .to_string(),
-        ))
+        )))
     }
 }
 
+async fn session_channel_store()
+-> pay_core::Result<(Arc<dyn pay_kit::mpp::store::ChannelStore>, bool)> {
+    channel_store(
+        &["PAY_MPP_REDIS_URL"],
```
EfeDurmaz16: This drops `PAY_SESSION_REDIS_URL` and `PAY_SESSION_REDIS_PREFIX`. Existing deployments using those settings will fall back to in-memory storage. Please keep the old names as fallbacks unless all deployments have already migrated.

rust/crates/mcp/src/tools/curl.rs
```diff
@@ -1081,6 +1089,77 @@ fn do_paid_fetch(
             );
             interpret_retry(fetch_request(&headers)?)
         }
+        RunOutcome::X402BatchChallenge { challenge, .. } => {
+            let built = pay_core::client::x402::build_batch_payment(
+                &challenge,
+                &store,
+                &session_cache.batch_channels,
+                None,
```
EfeDurmaz16, line 1097: Passing `None` funds only one request, so each later request needs an onchain top-up. The PR description says one deposit covers multiple MCP calls. Please update the description if this default is intentional (this way is also fully compatible with batch-settlement spec) ; a larger deposit option can follow separately.

rust/crates/pay-worker/src/bin/settle_sessions.rs
```diff
@@ -801,6 +1359,102 @@ struct SettlementCandidate {
...
+    if distribution_threshold_base_units.is_some_and(|threshold| {
+        distribution_threshold_reached(
+            state.settled_on_chain,
+            state.distributed_on_chain,
+            threshold,
+        )
+    }) {
+        return true;
+    }
+    let unsettled = state.cumulative > state.settled_on_chain;
```
EfeDurmaz16: If a fully settled channel is closed directly onchain, this check seems to skip it until Redis learns about the close. What picks it up if the client makes no further requests?

EfeDurmaz16 (commented):
Left a Redis compatibility note, a question about detecting onchain closes, and a non-blocking correction to the MCP description.

rust/crates/mcp/src/tools/curl.rs
lgalabru: Confirmed this default is intentional for now. I updated the PR description and in-code docs to state that MCP reuses the channel but currently tops up one request at a time; a configurable larger deposit can follow separately. This avoids claiming that the current default amortizes funding transactions.

rust/crates/cli/src/commands/server/start.rs
lgalabru: Fixed in dc2a89a6. `pay serve`, `settle-sessions`, and `close-channels` now keep `PAY_SESSION_REDIS_URL`, `PAY_SESSION_REDIS_PREFIX`, and `PAY_SESSION_FINALIZED_RETENTION_SECONDS` as fallbacks behind the new `PAY_MPP_*` names. The batch store also accepts the legacy session URL as its final shared-Redis fallback. Verified with `cargo check -p pay --features redis-session-store`.

rust/crates/pay-worker/src/bin/settle_sessions.rs
lgalabru: Fixed in dc2a89a6. The worker now treats an on-chain snapshot older than `PAY_X402_SNAPSHOT_MAX_AGE_SECS` (default 30s) as reconciliation-due even when Redis says the channel is fully settled. That periodic fetch observes a direct on-chain close, persists the refreshed watermark/check time, and drives finalize/distribute/reclaim without another client request. Added `batch_reconciliation_refreshes_stale_fully_settled_channels`; all 14 settle-sessions tests pass.

rust/crates/pay-worker/src/bin/settle_sessions.rs
```diff
@@ -334,56 +418,144 @@ impl SettlementRuntime {
...
+            let distribution_threshold_base_units =
+                parse_optional_positive_u64_env("PAY_X402_DISTRIBUTION_THRESHOLD_BASE_UNITS")?;
+            let settlement_max_idle = Duration::from_secs(parse_u64_env(
+                "PAY_X402_SETTLEMENT_MAX_IDLE_SECONDS",
+                DEFAULT_X402_SETTLEMENT_MAX_IDLE_SECONDS,
+            )?);
+            let snapshot_max_age = Duration::from_secs(parse_u64_env(
+                "PAY_X402_SNAPSHOT_MAX_AGE_SECS",
+                DEFAULT_X402_SNAPSHOT_MAX_AGE_SECONDS,
+            )?);
```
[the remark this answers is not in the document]
lgalabru: Fixed in 7432edcf. `PAY_X402_SNAPSHOT_MAX_AGE_SECS=0` now fails configuration loading with a clear greater-than-zero error before Redis/RPC reconciliation starts. Added `snapshot_max_age_must_be_positive`, documented the constraint, and verified all 15 settle-sessions tests plus pay-worker clippy.

EfeDurmaz16 (approved):

PR #462, opened 2026-09-11
feat(tx): adopt pay-kit's transaction path, Transaction V1 aware

lgalabru:
## Summary

Companion to solana-foundation/pay-kit#319 and solana-foundation/x402#11. Every place pay builds, signs, encodes, decodes or broadcasts a Solana transaction now goes through pay-kit's `core::tx`, `core::signing` and `core::rpc`. Legacy transactions are gone from pay: it builds version 0 everywhere, and version 1 (SIMD-0385) for operator-signed batches wherever the cluster gate is active. Stacked on `feat/x402-batch-settlement`; retarget to `main` once that lands.

## What changed

- **Dependencies**: pay-kit `1cfc850b`; solana-message ~4.4, solana-transaction ~4.1, solana-hash 4.4; surfpool main (the v1.4.0 tag predates V1); `[patch.crates-io]` on solana-keychain mirroring the kit workspace, to drop when keychain 2.0.0-beta.2 ships.
- **Ten sign-then-splice sites** (CLI gateway ATA creation, provider registration, subscription cancel direct and via gateway, pay-api redeem and subscription co-sign, operator plan broadcast, close-channels, bench fixtures and channel recovery) use the kit's slot signer and canonical (wincode) encoding. Blocking `RpcClient` sends go through `pay_kit::core::rpc`, since solana-rpc-client still serializes with bincode.
- **Push**: planner, permit and pay-api transfer batch work on `VersionedTransaction`, version 0 with the existing price-then-limit ComputeBudget prefix, so the wire is unchanged. The sponsor rejects legacy, lookup tables and any other version before validating the chunk. Both hand-mirrored chunk validators read the message through one static view.
- **Decoders** (transfer batch, subscription cancel, channel-open extraction, debugger correlation) accept versions 0 and 1 and reject legacy. `getTransaction` reads back with `maxSupportedTransactionVersion: 1`.
- **Servers** (Mpp, X402, upto, batch-settlement, session, subscription) are built with `with_tx_v1(Auto)`: one gate probe at startup, version 1 advertised and accepted when live. `settle-sessions`, `close-channels` and the session settlement worker pack at the probed version: 18 voucher settlements and 62 reclaims per transaction on V1 against 4 and 28.
- **Signers**: fee-payer signers are `Arc<dyn TransactionSigner>`. pay-api and pay-worker unwrap the keychain `Signer` enum into its Memory or GCP KMS backend, since the enum only signs messages.
- **Subscriptions**: adapted to pay-kit#316's wire rename (`planAddress`, `subscriptionProgram`, receipt `periodStart` / `periodIndex`; the receipt `reference` is the activation signature).

## Not in this PR

- `pay send --confidential` and pay-api's send endpoint delegate the confidential bundle to pay-kit; that bundle is on version 0 and still record-chunked. Collapsing it into one V1 transaction is the next kit change.
- Push chunks stay on version 0 (8 transfers). Moving them to V1 (~40 to 56 transfers) needs pay-api to advertise `transactionVersions` on the transfer-batch manifest.

## Verification

`cargo check --workspace --all-targets` and clippy clean; `cargo test --workspace` green, plus `cargo test -p pay-core --features server --test wrong_network_tests`.


rust/Cargo.toml
```diff
@@ -116,7 +116,7 @@ http-body = "1"
 # deps. The `axum` feature (kit's own gate) is intentionally OFF — pay has its
 # own gate (pay-core + pay-proxy).
 #
-pay-kit = { package = "solana-pay-kit", git = "https://github.com/solana-foundation/pay-kit.git", rev = "c143bfab", default-features = false, features = [
+pay-kit = { package = "solana-pay-kit", git = "https://github.com/solana-foundation/pay-kit.git", rev = "733fe68b", default-features = false, features = [
```
EfeDurmaz16, line 120: The updated pay-kit requires an authentication proof in subscription activation credentials, but `core/src/client/subscription.rs:340-352` still builds a transaction-only payload and wraps it directly in `PaymentCredential`. Both CLI and MCP use this builder, so their activations are rejected before broadcast.

Please use `build_subscription_activation_credential` or include the equivalent proof in the activation payload. The separate SIWMPP credential signed afterward does not satisfy this requirement.

rust/crates/core/src/client/subscription.rs
```diff
@@ -502,14 +502,14 @@ fn subscription_from_built_and_extensions(
 ) -> Subscription {
     Subscription {
         subscription_id: parsed.extensions.subscription_id.clone(),
```
EfeDurmaz16: The new receipt's `subscriptionId` is an opaque ID; the delegation PDA is now in `subscriptionDelegation`. This local field is still treated as a PDA by cancellation and authentication refresh, so callers of `persist_from_receipt` will persist an unusable address.

Please store `parsed.extensions.subscription_delegation` here and use different values for the two fields in the receipt fixture, which currently masks this mismatch.

This is limited to the public receipt helper: the built-in CLI/MCP paths currently derive the PDA separately.

rust/crates/core/src/server/subscription.rs
```diff
@@ -205,6 +206,7 @@ pub fn build_handler(
     };
 
     SubscriptionServer::new(config)
+        .map(|server| server.with_tx_v1(pay_kit::core::tx::TxV1Mode::Auto))
```
EfeDurmaz16, line 349: This handler passes `store: None`, and the gateway creates a fresh handler for each subscription request. Each gets a separate `MemoryStore`, so the proof binding saved during activation is lost before the next request.

A client using the SDK's proof flow can activate successfully, then receive `Subscription has no bound authentication proof` on its next request. Please share the store across requests, with persistence across restarts and instances where needed. The separate SIWMPP path is unaffected.

rust/crates/pay-worker/src/channel.rs
```diff
@@ -286,17 +286,16 @@ async fn find_open_signature(
...
 fn extract_open_ix_data(tx_b64: &str) -> Result<Vec<u8>, JobError> {
     use base64::Engine;
-    use solana_message::VersionedMessage;
     let raw = base64::engine::general_purpose::STANDARD
         .decode(tx_b64.trim())
         .map_err(|e| JobError::OpenIxDecode(format!("base64 decode: {e}")))?;
-    let tx: solana_transaction::versioned::VersionedTransaction = bincode::deserialize(&raw)
-        .map_err(|e| JobError::OpenIxDecode(format!("bincode deserialize: {e}")))?;
+    let tx = pay_kit::core::tx::decode_bytes(&raw)
+        .map_err(|e| JobError::OpenIxDecode(format!("transaction decode: {e}")))?;
```
EfeDurmaz16, line 293: `decode_bytes` rejects legacy transactions, but this function reads historical channel creation transactions. Existing channels opened with legacy transactions and nonempty distribution plans can no longer recover their recipients, blocking the distribution step in the settlement and closure workers.

Please preserve legacy decoding for historical reads, for example with canonical `wincode` deserialization into `VersionedTransaction`, while retaining the distribution-hash check. Empty distributions bypass this lookup.

EfeDurmaz16 (changes requested):
Found a few gaps around subscription activation, proof storage, and historical channel decoding. Left details inline, plus a smaller receipt-field mapping issue.

I’d address these before merging

rust/Cargo.toml
lgalabru: Fixed in 317bffd5: subscription activation now uses build_subscription_activation_credential, so the transaction and authentication proof are carried together in the credential used by CLI/MCP.

rust/crates/core/src/client/subscription.rs
lgalabru: Fixed in 317bffd5: persist_from_receipt stores extensions.subscription_delegation; the receipt fixture now uses distinct opaque subscription ID and delegation PDA values to prevent regression.

rust/crates/core/src/server/subscription.rs
lgalabru: Fixed in 317bffd5: gateway state now owns a shared subscription store, backed by the configured Redis session store for cross-request/restart/instance durability, with memory fallback for local operation.

rust/crates/pay-worker/src/channel.rs
lgalabru: Fixed in 317bffd5: historical channel reads use canonical wincode deserialization into VersionedTransaction, preserving both legacy and versioned transaction decoding while keeping the distribution-hash validation.

EfeDurmaz16 (approved):
Re-checked the latest changes. All four findings are addressed, and I found no further blockers. Thanks for the fixes!

PR #464, opened 2026-09-16
Pay Cloud M1: signing-backend registry, browser-linked remote wallet, Ledger

lgalabru:
Milestone 1 of the Pay Cloud plan (`docs/hosted-connector.md`): the keychain cleanup, a browser-linked "Remote wallet" setup flow with a local `pay-cloud` v0, an Openfort onboarding driver, and Ledger hardware wallets, verified on a device.

Rebased on `main` after #423 (Openfort remote signing) merged; pay-kit is pinned to its main head `0626cd7c` now that pay-kit #323 merged.

## Keychain: describe backends, don't match on them

- `pay_core::backend`: a `SigningBackend` trait (`custody`, `is_exportable`, `signs_raw_messages`, `approval`, `is_available`, `max_tx_version`, `deprecated`) and a registry. `--backend` flags, the picker, legacy signer sources and `Account::descriptor()` all resolve through it; there is no `match` on the storage kind outside `accounts.rs`, `signer.rs` and the account commands. `RemoteProvider` is a supertrait, so Circle later is one file plus a registry line. See `docs/keychain.md`.
- Export, approval gates and platform `cfg`s each live in one place. Every store exists on every OS and reports `is_available() == false` off-platform.
- 1Password is deprecated: existing accounts load, `pay account new` / `import` and the picker refuse it with the reason.

## Remote wallet onboarding (gh auth login pattern)

- `pay setup` shows the pay.sh banner and a richer picker: platform store, "Remote wallet (sign in from your browser)", and any hardware provider compiled in.
- The CLI binds a loopback `/callback`, opens `{cloud}/onboard?...` with PKCE S256, and exchanges the code for `{provider, credentials, wallet_id, pubkey}`; `register_provisioned_account` verifies the address with the provider and stores a normal remote account. `PAY_CLOUD_LOCAL=1` targets `http://127.0.0.1:8402`, `PAY_CLOUD_URL` any server, `PAY_NO_BROWSER=1` prints the URL instead.
- `crates/cloud` (`pay-cloud`): axum server embedding the `web-ui` cloud bundle (`pnpm build:cloud`, `include_dir`, same pattern as pdb). Sessions are in-memory, single-use, five minutes. Drivers implement `WalletDriver`; the Openfort driver (`openfort` cargo feature, default on) sends the browser to Openfort's own consent page, registers a fresh wallet secret, creates a Solana backend wallet and hands the CLI the project's credentials. pay owns no custody account; pay-cloud keeps nothing past the session.
- `web-ui/src/cloud`: a second Vite app. Opened from the CLI it renders as a full-screen terminal with the pay.sh wordmark; opened cold it shows the welcome card.

## Ledger

- `remote::ledger` behind the `ledger` cargo feature (pay-core and the CLI, opt-in because `hidapi` links IOKit/libudev). Hardware custody, not exportable, no raw messages, device confirmation, V0 cap. No credentials: `accounts.yml` holds the derivation path as the wallet id.
- The attributes do the work: `max_tx_version` is threaded into the charge, x402 exact/upto and channel-open builders; `require_raw_message_signing` refuses with the reason in the paths that need a raw signature (SIWMPP authenticate, subscription proofs, operator-signed session proofs, x402 sign-in, batch vouchers).
- Offer selection follows the signer too. A 402 often advertises an operator-signed session beside a flat charge (the Gemini gateway does). `RunOutcome::SessionChallenge` now carries the charge as `fallback`, and `RunOutcome::for_account` reads the paying account's backend from `accounts.yml` without prompting and swaps the fallback in when the backend cannot sign the session proof or an x402 sign-in. The CLI and the MCP `curl` tool both call it.
- Adopts pay-kit's transaction path (Transaction V1 aware) by cherry-picking #462 onto this branch; pinned to pay-kit main `0626cd7c`, solana-keychain `6461a18`.

## Verification

- Workspace `cargo fmt --check`, `cargo clippy --workspace --all-targets -D warnings`, all crate tests pass with and without `ledger` (pay-core 630/627, CLI 525, MCP 108); web-ui typecheck, build, build:cloud and 61 vitest tests pass.
- Loopback flow end to end against a local `pay-cloud` with a mock Openfort (consent, provision, exchange, account registered).
- Physical Ledger: `pay account new --backend ledger` reads the address from the device; a v0 MPP charge to debugger.pay.sh and a 0.045 USDC Gemini image generation on mainnet were both confirmed on the device, the latter after skipping the session offer.

## Milestone 2: the Grok connector (added 2026-09-16)

- **`/mcp`** in pay-cloud: rmcp streamable HTTP, a `PayMcp` per session, `Host` validation, bearer middleware. `PAY_CLOUD_MCP=1` mounts it with its OAuth server; `PAY_CLOUD_MCP_TOKENS` adds static tokens for hosts that only take a header.
- **OAuth 2.1 authorization server**: both well-known documents, DCR accepting Grok's exact registration, `/oauth/authorize` with mandatory S256 PKCE and RFC 8707 `resource`, a terminal-themed consent page, code and refresh grants (rotated), revocation. Opaque hashed tokens, 1 h access, 30 d refresh.
- **`PayContext` in pay-mcp**: every tool call resolves a `CallScope` (accounts store, overrides, approval, whether local files may be read). `LocalContext` reproduces `pay mcp`; pay-cloud's `CloudContext` resolves the bearer's tenant. Remote credentials follow a `CredentialSource` on `AccountsStore`, so a per-tenant in-memory store signs through the unchanged payment paths.
- **Spending policy as approval**: payment intents carry their exact amount; `PolicyGate` enforces a per-call ceiling and daily cap, records at approval, and chains the client's elicitation when it has one.
- **Consent creates the wallet**: a browser with no wallet is offered the providers; the Openfort hop provisions one, binds a tenant, approves the request, and a `pay_subject` cookie lets the same browser reuse its wallet for the next client.
- **Identity**: a tenant's subject is derived from the provider account (Openfort project id), so a returning user gets the same wallet from any browser; the cookie is only a shortcut and a forged one is refused. The OAuth store takes a clock and expiry is tested on schedule.
- **Hosts registry** (`hosts.rs`): Grok, Claude, ChatGPT/Codex, Cursor, Claude Code and Codex CLI as OAuth clients, each with its redirect URIs and quirks; redirect policy and consent naming follow from it.
- **Grok, live**: connects and lists tools through a Tailscale Funnel hostname. Its custom connector sends no header and abandons OAuth after registration, so the demo runs a dev-only anonymous tenant (`run.sh --funnel --anonymous`); the plan doc records the request log for xAI.
- Not in this PR: persisting tenants and OAuth state (Postgres, encrypted credentials), a `/connect` page for limits and revocation, Redis MCP sessions, deployment, and the conformance run against Grok's real connector UI, which needs a public URL.

Also here: card purchases of USDC through Coinflow (`/fund` page, `POST /api/fund/*`, `PAY_ONRAMP=coinflow` in the TUI), behind merchant settings Coinflow still has to enable. MoonPay stays the default until then.

## Not in this PR (milestone 1)

- Real-Openfort run of the consent/provision flow (verified against a mock only; the three open questions are in the plan doc), and the connector custody decision.
- `ledger` in the Homebrew/npm release builds (Linux needs `libudev-dev`), the envelope-aware `signature_type` at the spec level, and two keychain-side rough edges seen on the device run: `solana-remote-wallet` prints its approval prompt to stdout, and its Trezor-bridge probe logs connection-refused noise.


mehdinemati8889-dotcom (approved):

mehdinemati8889-dotcom (commented):
عتایببببب

rust/crates/cli/src/commands/account/new.rs
```diff
@@ -22,15 +23,28 @@ pub struct NewCommand {
     /// Replace existing account.
     #[arg(long)]
     pub force: bool,
+
+    /// Remote-backend credential as `key=value`, repeatable (e.g.
+    /// `--credential secret_key=sk_live_…`). Each field also reads
+    /// `{PROVIDER}_{FIELD}` from the environment.
+    #[arg(long = "credential", value_name = "KEY=VALUE")]
+    pub credentials: Vec<String>,
```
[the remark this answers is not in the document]
lgalabru: Fixed in 30a7d2ef: `--credential` now takes `key=@file` and reads the value from the file; an inline value is refused, pointing at the file form or the existing `{PROVIDER}_{FIELD}` env var. Errors never echo the flag. Applies to both `pay account new` and `pay setup`.

rust/crates/cloud/src/onboard.rs
```diff
@@ -0,0 +1,687 @@
...
+
+    let wallet = driver
+        .provision(&grant)
+        .await
+        .map_err(ApiError::from_driver)?;
+    let redirect = session.redirect_url();
+    let address = wallet.address.clone();
+    tracing::info!(provider = driver.id(), address = %address, "wallet provisioned");
+    if !state.attach_wallet(&echoed_state, wallet) {
+        return Err(ApiError::invalid_grant());
```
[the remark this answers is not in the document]
lgalabru: Fixed in 9a93d265: `AppState::claim_provisioning` reserves the session under the lock before the driver runs. A concurrent completion gets 409 `provisioning`, a later one `already_completed`, and a failed provisioning releases the claim so a retry works. Covered by a test with a gated driver: two concurrent completions, one provider call.

rust/crates/cli/src/commands/cloud_onboard.rs
```diff
@@ -0,0 +1,867 @@
...
+pub fn run_setup_onboarding(account: &str, force: bool) -> pay_core::Result<()> {
+    let result = run_loopback_onboarding(&OnboardRequest {
+        cloud_url: default_cloud_url(),
+        account: account.to_string(),
+    })?;
+    if result.is_ready() {
+        return register_provisioned_account(account, &result, force);
+    }
+    print_result(&result);
+    Ok(())
```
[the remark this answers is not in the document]
lgalabru: Fixed in c50c9a62: anything but `status: ready` is now an error naming the status and the server message, and says nothing was saved.

rust/crates/cloud/src/lib.rs
```diff
@@ -0,0 +1,599 @@
...
+
+    /// Store a session, purging expired ones first.
+    pub fn insert_session(&self, session: OnboardSession) {
+        let now = Instant::now();
+        let mut sessions = self.sessions.lock().unwrap();
+        let mut by_state = self.by_state.lock().unwrap();
+        sessions.retain(|_, s| !s.is_expired_at(now));
+        by_state.retain(|_, key| sessions.contains_key(key));
+        by_state.insert(session.state.clone(), session.key());
+        sessions.insert(session.key(), session);
```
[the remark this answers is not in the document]
lgalabru: Partly addressed in 9a93d265: the store is bounded at 4096 sessions. At capacity it sweeps expired entries and otherwise refuses with 503 `busy`; live sessions are never evicted, and the per-insert full-map sweep is gone (expired entries are invisible to lookups until swept). Per-client rate limiting is left to the edge in front of pay-cloud rather than done in-process.

rust/crates/pay-worker/src/bin/settle_sessions.rs
```diff
@@ -1112,8 +1112,22 @@ async fn run_batch(runtime: &SettlementRuntime) -> Result<SettleSessionsMetrics,
...
+                    .resolve(&pay_kit::mpp::solana_rpc_client::rpc_client::RpcClient::new(rpc_url)),
+            )
+        })
+        .await
+        .unwrap_or(pay_kit::core::tx::TxVersion::V0)
+    };
     let handle = spawn(
-        SettlementConfig::new(runtime.operator, Arc::clone(&runtime.signer)),
+        SettlementConfig::new(runtime.operator, Arc::clone(&runtime.signer))
+            .with_tx_version(tx_version),
```
EfeDurmaz16: V1 is already active on mainnet and devnet, so this selects it today. The pinned packer checks bytes but misses the 64-account limit: I reproduced a 15-payer distribution batch with 67 accounts and 32 reclaims with 66 accounts, both within the byte limit but rejected by `build_unsigned`. Could we keep both workers on V0 until the packer enforces the account limit too?

rust/crates/core/src/backend/local.rs
```diff
@@ -0,0 +1,537 @@
...
+impl LocalKeystoreBackend for File {
+    fn kind(&self) -> BackendKind {
+        BackendKind::File
+    }
+
+    fn keystore(&self, params: &StoreParams<'_>, gate: Gate) -> Result<Keystore> {
+        let path = params
+            .file_path
+            .ok_or_else(|| Error::Config("The file backend needs a keypair path.".to_string()))?;
+        let store = crate::keystore::store::FileStore::new(path);
```
EfeDurmaz16, line 359: This changes how existing file-backed accounts load. The previous loader expanded `~` and accepted base58 keypair files; `FileStore` does neither. I reproduced both failures through the registered-account path. Could we preserve those formats here?

rust/crates/cli/src/commands/account/import.rs
```diff
@@ -77,8 +76,8 @@ impl ImportCommand {
             None => super::new::pick_backend()?,
         };
 
-        let (ks, keystore_kind, _) =
-            super::import::build_keystore(&backend_id, self.vault.as_deref(), &name)?;
+        let (ks, keystore_kind, _, _) =
+            super::new::build_keystore(&backend_id, self.vault.as_deref(), &name)?;
 
         if backend_id == "file" && ks.exists(&name) {
```
EfeDurmaz16: With biometrics unavailable, this can now replace an existing Keychain entry without confirmation when stderr is not a TTY or the name is missing from the mainnet account map. The existence check below only covers files. Could import require explicit overwrite authorization for existing entries, as `account new` does?

EfeDurmaz16 (changes requested):
Thanks! I verified this at `67fe657` and reproduced the three regressions noted inline. I kept the pay-cloud security findings out of the public review.

rust/crates/cloud/src/drivers/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+    fn parse_grant(&self, fragment: &str) -> Result<(ConsentGrant, String), DriverError>;
+
+    /// Turn a grant into a wallet the CLI can use.
+    async fn provision(&self, grant: &ConsentGrant) -> Result<ProvisionedWallet, DriverError>;
+
+    /// What identifies the user's account at the provider across sign-ins,
+    /// when the grant carries it: the project id for Openfort. A returning
+    /// user is recognised by this, not by a cookie, so a wallet survives a
+    /// lost browser.
+    fn account_identity(&self, grant: &ConsentGrant) -> Option<String> {
```
EfeDurmaz16: `account_identity` trusts `project_id` from the returned fragment, so an attacker can complete their own OAuth/PKCE flow with a victim project ID, overwrite that tenant's API key, and receive its session. Please resolve the tenant from a server-verified provider identity before refreshing credentials or issuing a code.

rust/crates/cloud/src/oauth.rs
```diff
@@ -0,0 +1,2965 @@
...
+        })
+        .into_response());
+    }
+    let (subject, fresh_login, bound) = match bearer {
+        Some(token) => {
+            let (subject, bound) = approve_with_privy(&state, token).await?;
+            (subject, true, bound)
+        }
+        None => (
+            crate::tenants::cookie::subject(&headers)
```
EfeDurmaz16: `pay_subject` is unsigned and accepted as authorization here. Anyone who knows a subject can send it directly in `Cookie` and approve OAuth as that tenant. Please use an opaque server-side session or an authenticated, expiring cookie.

rust/crates/cloud/src/oauth.rs
```diff
@@ -0,0 +1,2965 @@
...
+    // every error, so a provider failure can be retried safely.
+    let claim = state
+        .tenants()
+        .claim_link(&ticket)
+        .ok_or_else(unknown_link)?;
+    let guest = claim.subject().to_string();
+    let (subject, bound) = approve_with_privy(&state, token).await?;
+    let record = state.tenants().get(&subject).ok_or_else(unknown_request)?;
+    let mut linked = (*record).clone();
+    linked.subject = guest.clone();
```
EfeDurmaz16, line 1384: Link tickets are transferable bearer tokens. A victim who opens an attacker's link and signs in has their wallet bound to the attacker's guest subject and existing tokens. Please bind the ticket to the initiating browser/session and require explicit confirmation.

rust/crates/core/src/client/subscription.rs
```diff
@@ -335,6 +335,8 @@ pub fn build_credential_with_authenticate_and_override(
         }
     }
 
+    // Activation binds a reusable bearer proof signed as a raw message.
+    signer.require_raw_message_signing("a subscription proof")?;
```
EfeDurmaz16, line 350: Policy uses outer `request.amount`, while the signed activation uses `methodDetails.amount`; I reproduced a $1 policy approval signing a $100 transfer. Later pulls also bypass `PolicyGate`. Please enforce one canonical amount and the recurring cap/cadence at every pull.

EfeDurmaz16 (commented):
Four pay-cloud security items to fix before deployment are inline.

rust/crates/pay-worker/src/bin/settle_sessions.rs
lgalabru: Fixed in 1f1cf020 without rolling either worker back to V0. Reclaim batches now compile every proposed batch through build_unsigned, so byte, static-account, instruction, and signature limits all participate in packing; a regression test covers 62 V1 reclaims with distinct rent payers. Distribution settlement remains V1 and is conservatively capped at 14 operations for this heterogeneous workload, below the reproduced 15-operation/67-account case.

rust/crates/core/src/backend/local.rs
lgalabru: Fixed in 1f1cf020. Registered file backends expand tilde paths before constructing FileStore, and FileStore now accepts both 64-byte JSON arrays and base58 keypair contents. Added a base58 load/derived-pubkey regression test; pay account import accepts both formats too.

rust/crates/cli/src/commands/account/import.rs
lgalabru: Fixed in 1f1cf020. pay account import now requires explicit --force whenever either the account mapping or the selected keystore backend already contains that name, including Keychain and headless cases. The old TTY-dependent overwrite path and file-only existence check are gone.

rust/crates/cloud/src/drivers/mod.rs
lgalabru: Fixed in 1f1cf020 and hardened further in d7272a36. project_id is treated only as untrusted display metadata and cannot select a tenant. Openfort binds its tenant to possession of the high-entropy API credential via HMAC-SHA256 under a random server key; the raw credential is used only transiently and is never stored as identity. A different key claiming the victim project ID gets a different subject and cannot refresh or approve as the victim. Regression tests cover identical-key/different-project and different-key/same-project cases.

rust/crates/cloud/src/oauth.rs
lgalabru: Fixed in 1f1cf020. pay_subject now carries an HMAC-SHA256 signature under a per-registry random server key and is verified in constant time before use. Unsigned, tampered, and cookies minted by another registry are rejected; Max-Age, Secure, HttpOnly, and SameSite attributes remain in place.

rust/crates/cloud/src/oauth.rs
lgalabru: Fixed in 1f1cf020. GET of the live link now issues a signed, ticket-bound same-browser cookie containing only the ticket hash, and POST requires both that cookie and an explicit confirm=true JSON payload before it reserves the ticket or contacts Privy. Tests cover missing auth, missing confirmation, and the successful bound flow.

rust/crates/core/src/client/subscription.rs
lgalabru: Fixed in 1f1cf020. Decode now requires exact equality between outer request.amount and signed methodDetails.amount, and between requested cadence and expectedPeriodHours, before policy or signing. For hosted PolicyGate, recurring authorization is now rejected fail-closed because later merchant-driven on-chain pulls do not traverse this process and therefore cannot honestly be recorded in its in-memory daily ledger. Non-hosted subscriptions remain governed by the signed on-chain plan amount and cadence. Regression tests cover amount mismatch, cadence mismatch, and hosted recurring rejection.

EfeDurmaz16 (commented):
One V1 batching blocker remains inline; the other requested fixes look good.

rust/crates/pay-worker/src/bin/settle_sessions.rs
```diff
@@ -1112,8 +1112,31 @@ async fn run_batch(runtime: &SettlementRuntime) -> Result<SettleSessionsMetrics,
...
+    };
+    let mut settlement_config =
+        SettlementConfig::new(runtime.operator, Arc::clone(&runtime.signer))
+            .with_tx_version(tx_version);
+    if tx_version == pay_kit::core::tx::TxVersion::V1 {
+        // Fifteen distribution settlements can introduce 67 distinct static
+        // accounts. Keep V1 enabled, but cap this heterogeneous workload below
+        // its 64-account ceiling until pay-kit packs against compiled messages.
+        settlement_config.max_voucher_settlements_per_tx =
+            settlement_config.max_voucher_settlements_per_tx.min(14);
```
EfeDurmaz16, line 1136: Correction: my two-recipient-per-channel repro was not valid for batch-settlement SVM. The scheme requires exactly one `payTo` recipient at 10,000 bps, and the pinned facilitator rejects any other distribution. With the required shared `payTo`, 14 distributions compile at exactly 64 accounts, so this is not a blocker. Mixed `payTo` channels sharing one worker/Redis prefix could still exceed the cap; that can be handled as a topology or packing follow-up.

EfeDurmaz16 (approved):
LGTM

